PR #1, opened 2019-04-23
Add driven controls

virginia-m:
Adding driven controls to the library. Includes several predefined driven controls + tests and an example notebook.


michaelhush:
Few more changes.

The module name is driven controls (this is fine because it contains multiple driven controls).

However the class name should be DrivenControl (not DrivenContorls) furthermore the file containing DrivenControl should also be named DrivenControl.

The names of the driven controls would be better if we use the common abbreviation, similar to BLACK OPAL. This is within our conventions as most of the DrivenContorl names are longer than three words. 

'wimperis_1' => 'BB1'

'solovay_kitaev_1' => 'SK1'

'walsh_amplitude_modulated_filter_1' => 'WAMF1'

'compensating_for_off_resonance_with_a_pulse_sequence' => 'CORPSE'

'compensating_for_off_resonance_with_a_pulse_sequence_with_wimperis' => 'CORPSE in BB1'

'compensating_for_off_resonance_with_a_pulse_sequence_with_solovay_kitaev' => 'CORPSE in SK1'

'short_composite_rotation_for_undoing_length_over_and_under_shoot' => 'SCROFULOUS'

'corpse_in_scrofulous_pulse' => 'CORPSE in SCROFULOUS'

Note in all cases the variable name should be the capitalized version of the same words. So for the last one it would be:

CORPSE_IN_SCROFULOUS = 'CORPSE in SCROFULOUS'

michaelhush:
The get_plot_formatted_arrays method should return a dictionary rather than a tuple to match the dynamical decoupling sequence pattern.

The name of the arrays should change depending on the coordinates.

'cylindrical'
rabi_rate
azimuthal_angle
detuning

'cartesian'
amplitude_x
amplitude_y
detuning

michaelhush:
The keyword dimensionless should be changed to dimensionless_rabi_rate throughout the driven controls module.

michaelhush:
The driven controls constructor need to be brought inline with the constructor of the dynamical decoupling sequence.

Instead of simply providing segments, four arrays should be passed:

rabi_rates
azimuthal_angles
detunings
durations

The following defaults and checks should occurs:

* Check all the Rabi rates are greater or equal to zero
* Check all of these arrays are the same length.
* If the rabi_rates, azimuthal_angles or detunings are not provided i.e. are None then they should be created with zeros for all values with the same length as the others.
* If durations is not provided it i.e. None then it should be create with ones for all the values and be the same length as the others

This is now in the "cylindrical" coordinates and provided with a set of meaningfully named arrays, which brings driven controls in line with the dynamical decoupling sequences. 

In terms of attributes for the driven controls class:

* self.segments should no longer be an attribute (and shouldn't be anywhere anymore).
* self.number_of_segments should remain an attribute.
* self.amplitudes doesn't seem to be used anywhere and should be removed.
* self.segment_durations should be removed (it is now covered by self.durations in the constructor).
* self.angles should be a property not an attribute.
* self. directions should be a property not an attribute.
* self.segment_times should be renamed to self.times and should be a property.
* self.maximum_amplitude should be removed. 
* self.maximum_rabi_rate  should be a property.
* self.maximum_detuning should be a property.
* self.minimum_duration should be a property.
* self.amplitudes_x should be added as a property.
* self.amplitudes_y should be added a property.

With the deletion of segments, and the changes above, get_transformed_segments no longer adds much value. It should be deleted. As far as I can tell it is only used in one part, and it's functionality will be largely replaced with all the properties that are defined above.  

The pretty print str also needs to be updated to display the

rabi_rates
azimuthal_angles
detunings
durations

Also I made a mistake WRT to the plotting function, I forgot the plural forms for some of the words, these need to be changed to:

rabi_rate -> rabi_rates
azimuthal_angle -> azimuthal_angles
detuning -> detunings
amplitude_x -> amplitudes_x
amplitude_y -> amplitudes_y

virginia-m:
Hi Rajib,
Can you please change the constructor such that it explicitly converts any input to a numpy array? Then the input can be a Python list or a numpy array.

PR #5, opened 2019-05-23
Adding cirq support

rajibchakravorty:
Major : Add method to convert a DDS to a `cirq.Circuit` or `cirq.Schedule`

Request for review:
1. `circuit_type` is an option with 2 possible string values - a) 'standard circuit' - where the circuit will contain sequence operations and gaps will be filled up by identity gates; returns a `cirq.Circuit` instance b) 'scheduled circuit' - where sequence operations are scheduled to be occurring at offsets and does not require any Identity gates to fill; returns a `cirq.Schedule` instance; in either case the returned instance is accepted by `cirq.Simulator` for simulation (both are illustrated in the notebook. Please check if the string keys make sense.

2. Example notebook in examples/run_a_dds_on_cirq.ipynb.

Other additions/points to consider:

1. Pre-Post gates are specified here as a . 2x2 unitary matrix. This can accommodate the default X_{pi/2} or other rotations too (similar to angle notation in U3 gate in Qiskit)
2. The drawing of circuit is not great in cirq. Especially the 2x2 matrix at both ends looks bad (see the example notebook). But I could not find a nicer way to draw the circuit. Same comment applies to result - no nice direct API to draw/beatify the outcome.
3. In the absence of an actual device, demonstration of noisy system was not possible [so removed the comparison with Ramsey sequence]. Alternative was to use artificial noise models. But the library provides limited noisy gates/channels and the knowledge of noise had to known a-priori. I would consider this as beyond scope of the notebook at this moment.


qctrlopencontrols/cirq/cirq_circuit.py
```diff
@@ -0,0 +1,505 @@
...
+                number_of_id_gates += 1
+            circuit_operations.append('offset')
+            time_covered = offsets[operation_idx] + unitary_time
+
+    return circuit_operations
+
+
+def _get_rotations(operation):
+
+    """Returns the pulses based of the rotation operation
```
charmasaur: of -> on?

qctrlopencontrols/cirq/cirq_circuit.py
```diff
@@ -0,0 +1,505 @@
...
+    Parameters
+    ----------
+    operation : numpy.ndarray
+        1-D array (length=3) consisting of rabi rotation, azimuthal_angle
+        and detuning_rotation at an offset of a sequence
+
+    Returns
+    -------
+    numpy.ndarray
+        A 1-D array of length 3 containing x_rotation, y_rotation and z-rotation
```
charmasaur: z-rotation -> z_rotation

qctrlopencontrols/cirq/cirq_circuit.py
```diff
@@ -0,0 +1,505 @@
...
+        calculate from sequence operation
```
charmasaur: calculated*

qctrlopencontrols/cirq/cirq_circuit.py
```diff
@@ -0,0 +1,505 @@
...
+        A 2x2 unitary matrix as pre-post gate operations. Defaults to
+        the unitary matrix corresponding to a rotation of :math:'\\pi/2' around
+        X-axis. If None, pre-post gate is omitted from the circuit.
+    add_measurement : bool, optional
+        If True, the circuit contains a measurement operation for each of the
+        target qubits. Measurement from each of the qubits is associated
+        with a string as key. The string is formatted as 'qubit-X' where
+        X is a number between 0 and len(target_qubits).
+    circuit_type : str, optional
+        One of 'scheduled circuit' or 'standard circuit'. In the case of
```
charmasaur: I'm not really sure about python best practice here, but instead you could refer to constants.SCHEDULED_CIRCUIT and constants.STANDARD_CIRCUIT. That might make it a bit harder to pass an invalid value (even though we don't have a compiler, people might use IDEs that warn them about using non-existent constants).

qctrlopencontrols/cirq/cirq_circuit.py
rajibchakravorty: Addressed

qctrlopencontrols/cirq/cirq_circuit.py
rajibchakravorty: Addressed

qctrlopencontrols/cirq/cirq_circuit.py
rajibchakravorty: Addressed

qctrlopencontrols/cirq/cirq_circuit.py
rajibchakravorty: I went with legacy here. If this needs to be changes, we have to check carefully other similar instances.

rajibchakravorty:
Updates pushed.

1. `pre_post_rotation` is removed from both DDS constructor and circuit construction; added to `predefined` sequences as boolean option (if True adds a X_{pi/2} pulse on both ends of the sequence. Default is `False` - justification is the sequences are not defined with rotations at both ends - it is only added for some specific experiments; consequence is that the notebooks to 'run_a_dds_on_xxxx' have changes in where we create the sequences in the beginning (add a `pre_post_rotation=True` argument). Otherwise they are same.

2. Circuit construction as a result became more straightforward. Both Qiskit and Cirq conversion methods are updated due to this.

michaelhush:
Comments:

1. There is some inconsistency with the naming. Currently we use IBM Q and cirq in the same way which isn't quite right. I think we should change the example file, and accordingly title, for `running_a_dds_on_ibm_q.ipynb` to `export_a_dds_to_qiskit.ipynb`. This actually matches the description of the file in the readme, so that helps. This also means we should change cirq example to `export_a_dds_to_cirq.ipynb`.

2. Add a brief description of the Export a DDS to Cirq jupytet notebook to the README.

rajibchakravorty:
- Notebook names and titles changed
- Add a couple of sentences on Cirq notebook in Readme.

michaelhush:
Consistency comments:

With qiskit, we have a `qiskit` module then we have the method `convert_dds_to_quantum_circuit` which returns a `QuantumCircuit` object. The file is called `quantum_circuit.py`. I want to match this pattern as much as possible.

Cirq currently has a method `convert_dds_to_cirq_circuit` which returns two different types of objects, Circuits or Schedules. It also has the term cirq inserted. This breaks the pattern above. 

To change the cirq module to match the qiskit module we should remove this "circuit_type" keyword altogether (and it's constants) and instead use two method:

`convert_dds_to_circuit`
`convert_dds_to_schedule`

Each method now returns a predictable type, which is clearer. They are implied to be for cirq through the module. Technically they should also be in separate files `circuit.py` and `schedule.py` to keep the pattern consistent. If there is any shared code I wouldn't mind it being in one of them and imported by the other. But just make sure there is not mutual importing errors.

rajibchakravorty:
updates:

1. Conversion to `cirq` `circuit` and `schedule` separated in two files (with common methods imported from `qiskit` module where needed
2. Notebooks updated accordingly. Minor change in qiskit notebook also (e.g. matching the gate_time in description and actual code and some tex format for micorseconds etc.).

qctrlopencontrols/cirq/circuit.py
```diff
@@ -0,0 +1,225 @@
...
+"""
+
+import numpy as np
+
+import cirq
+
+from qctrlopencontrols.dynamic_decoupling_sequences import DynamicDecouplingSequence
+from qctrlopencontrols.exceptions import ArgumentsValueError
+
+from qctrlopencontrols.qiskit import (
```
michaelhush: cirq should not import anything from qiskit. Could you pull these constants into global and put the method in a module where it would be more appropriate.

rajibchakravorty:
@michaelhush 

- Troubling methods are removed - so no cross-module imports necessary anymore

PR #6, opened 2019-05-30
Remove get plot method

rajibchakravorty:
Major modification - public method get_plot_data_from_segments removed and its functions are absorbed in the get_plot_formatted_arrays method

Minor modification - convert_dds_to_driven_controls->convert_dds_to_driven_control and changes related to this.

### DIFF
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -533,27 +533,27 @@ class DrivenControl(QctrlObject):   #pylint: disable=too-few-public-methods
         plot_amplitude_z = np.concatenate(
             ([0.], (plot_amplitude_z[:, np.newaxis] * np.ones((1, 2))).flatten(), [0.]))
 
-        plot_dictionary = {
-            'amplitudes_x': plot_amplitude_x,
-            'amplitudes_y': plot_amplitude_y,
-            'detunings': plot_amplitude_z,
-            'times': plot_time
-        }
+        plot_dictionary = {}
+        if coordinates == CARTESIAN:
+            plot_dictionary = {
+                'amplitudes_x': plot_amplitude_x,
+                'amplitudes_y': plot_amplitude_y,
+                'detunings': plot_amplitude_z,
+                'times': plot_time}
 
-        if coordinates == CYLINDRICAL:
+        elif coordinates == CYLINDRICAL:
             x_plot = plot_amplitude_x
             y_plot = plot_amplitude_y
             x_plot[np.equal(x_plot, -0.0)] = 0.
             y_plot[np.equal(y_plot, -0.0)] = 0.
             azimuthal_angles_plot = np.arctan2(y_plot, x_plot)
             amplitudes_plot = np.sqrt(np.abs(x_plot**2 + y_plot**2))
+
             plot_dictionary = {
                 'rabi_rates': amplitudes_plot,
                 'azimuthal_angles': azimuthal_angles_plot,
                 'detunings': plot_amplitude_z,
-                'times': plot_time
-            }
-
+                'times': plot_time}
         return plot_dictionary
 
     def __str__(self):
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -508,39 +508,30 @@ class DrivenControl(QctrlObject):   #pylint: disable=too-few-public-methods
             normalizer = 1
 
         if coordinates == CARTESIAN:
-            segments = np.vstack((self.amplitude_x / normalizer,
-                                  self.amplitude_y / normalizer,
-                                  self.detunings,
-                                  self.durations)).T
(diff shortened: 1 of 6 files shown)


rajibchakravorty:
Closing it now; The task is not complete.

rajibchakravorty:
- Loop removed from get_plot_method in Driven Controls

qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -31,54 +31,6 @@
     UPPER_BOUND_DURATION, LOWER_BOUND_DURATION)
 
 
-def get_plot_data_from_segments(segments):
```
charmasaur: By deleting this we're changing the API in a backwards-incompatible way, so could potentially break other packages that depend on open controls. I'm not sure how releases of the package are handled -- do we need a version bump or something?

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -95,7 +95,7 @@ def _check_maximum_rotation_rate(
                     'allowed_maximum_detuning_rate': UPPER_BOUND_DETUNING_RATE})
 
 
-def convert_dds_to_driven_controls(
+def convert_dds_to_driven_control(
```
charmasaur: Same story here -- if an external package is using the old method, they'll now be broken (which might be fine, I'm just not sure).

qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -545,42 +497,62 @@ def get_plot_formatted_arrays(self, coordinates=CARTESIAN, dimensionless_rabi_ra
...
+        plot_amplitude_z = control_segments[:, 2]
+
+        plot_amplitude_x = np.concatenate(
+            ([0.], (plot_amplitude_x[:, np.newaxis] * np.ones((1, 2))).flatten(), [0.]))
+        plot_amplitude_y = np.concatenate(
+            ([0.], (plot_amplitude_y[:, np.newaxis] * np.ones((1, 2))).flatten(), [0.]))
+        plot_amplitude_z = np.concatenate(
+            ([0.], (plot_amplitude_z[:, np.newaxis] * np.ones((1, 2))).flatten(), [0.]))
+
+        plot_dictionary = {
```
charmasaur: IMO it'd be a bit clearer to put this in an "if coordinates == CARTESIAN" block, so we're only setting plot_dictionary once (alternatively, we could just return the dictionary directly from the two if blocks, there's no need for the local variable at all).

qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -545,42 +497,62 @@ def get_plot_formatted_arrays(self, coordinates=CARTESIAN, dimensionless_rabi_ra
...
-                           self.detunings, self.durations)).T
-            )
+            control_segments = np.vstack((
+                self.rabi_rates / normalizer,
+                self.azimuthal_angles,
+                self.detunings,
+                self.durations)).T
+
+        segment_times = np.insert(np.cumsum(control_segments[:, 3]), 0, 0.)
+        plot_time = (segment_times[:, np.newaxis] * np.ones((1, 2))).flatten()
```
charmasaur: optional: since this trick for duplicating the arrays is used in a few places, and isn't super intuitive (to me, anyway) you could pull it out into a little private helper function (which could even be defined inside this method). That lets you save some code, and implicitly gives a little bit of documentation (in the function name).

qctrlopencontrols/driven_controls/driven_control.py
rajibchakravorty: Actually this method was never used anywhere and hence should be removed anyway.

charmasaur: Not used anywhere by us, but what if there's a third party package depending on it?

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
rajibchakravorty: This is a valid concern. This method was public. IMHO - we should be more consistent in naming from the start.

qctrlopencontrols/driven_controls/driven_control.py
rajibchakravorty: @charmasaur - I agree. Linter objects to some other ways. Just updated a method that may be a bit clearer and linter friendly.

qctrlopencontrols/driven_controls/driven_control.py
rajibchakravorty: This was a helper function to another method that actually gets the data for plotting. So the user should use this other method instead. I am pretty sure this should not be a problem.

charmasaur: OK, sounds good. As this package matures I expect we'll need to introduce a proper versioning scheme, so that we can have proper deprecation policies for this sort of change, but in these early days I can't imagine too much going wrong.

PR #7, opened 2019-06-04
Update version.py

michaelhush:
Updated the version number.

### DIFF
--- a/qctrlopencontrols/version.py
+++ b/qctrlopencontrols/version.py
@@ -18,4 +18,4 @@ Q-CTRL Open Controls
 ===========================
 """
 
-__version__ = '1.1.1'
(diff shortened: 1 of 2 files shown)


charmasaur:
Out of curiosity, what's the version number schema?

qctrlopencontrols/version.py
```diff
@@ -18,4 +18,4 @@
 ===========================
 """
 
-__version__ = '1.1.1'
+__version__ = '1.2.1'
```
charmasaur: Just saw the email, should this be 1.2.0?

michaelhush: Urgh, since I messed up it has to become 1.2.1

PR #8, opened 2019-06-11
Defining the public API

rajibchakravorty:
Major changes

- `QCtrlObject` removed from package
- __repr__ is carried out by method `create_repr_from_attributes(..)` in `base.utils` module
- `__all__` introduced in package and sub-package
- public methods for creating specific DDS and Driven Control are changed to private methods; only `new_predefined_dds` or `new_predefined_driven_control` are public
- `qiskit` package method renamed to `covert_dd_to_qiskit_quantum_circuit`


qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -594,6 +602,29 @@ def __str__(self):
...
+
+        attributes = {
+            'rabi_rates': self.rabi_rates,
+            'azimuthal_angles': self.azimuthal_angles,
+            'detunings': self.detunings,
+            'durations': self.durations,
+            'name': self.name
+        }
+
+        class_name = '{0.__class__.__name__!s}'.format(self)
```
charmasaur: Rather than requiring each class do this manually, could we change create_repr_from_attributes to take the object itself and do this formatting internally?

qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
```diff
@@ -72,14 +72,15 @@ def __init__(self,
                  detuning_rotations=None,
                  name=None
                  ):
-
+        """
```
charmasaur: Delete this block

qctrlopencontrols/driven_controls/__init__.py
```diff
@@ -31,12 +31,11 @@
...
-    new_primitive_control, new_wimperis_1_control, new_solovay_kitaev_1_control,
-    new_compensating_for_off_resonance_with_a_pulse_sequence_control,
-    new_compensating_for_off_resonance_with_a_pulse_sequence_with_solovay_kitaev_control,
-    new_compensating_for_off_resonance_with_a_pulse_sequence_with_wimperis_control,
-    new_short_composite_rotation_for_undoing_length_over_and_under_shoot_control,
-    new_walsh_amplitude_modulated_filter_1_control,
-    new_corpse_in_scrofulous_control)
+from .predefined import new_predefined_driven_control
+
+__all__ = ['UPPER_BOUND_RABI_RATE', 'UPPER_BOUND_DETUNING_RATE',
```
charmasaur: IMO it'd be best to have one item per line, in alphabetical order. Makes scanning the code a little easier.

charmasaur: I can't remember what we decided for __all__ in subpackages -- did we decide it was necessary, or just that the API of qctrlopencontrols is just the contents of the top-level __all__?

qctrlopencontrols/__init__.py
```diff
@@ -14,14 +14,21 @@
...
 
-from .dynamic_decoupling_sequences import (DynamicDecouplingSequence,
-                                           new_predefined_dds,
-                                           convert_dds_to_driven_control)
-from .driven_controls import DrivenControl, new_predefined_driven_control
-from .qiskit import convert_dds_to_quantum_circuit
-from .cirq import (convert_dds_to_cirq_circuit,
-                   convert_dds_to_cirq_schedule)
+from . import dynamic_decoupling_sequences
+from .dynamic_decoupling_sequences import *
```
charmasaur: I'm not sure about this -- it means the user has to look into the sub-packages to see the API. Although I suppose they have to do that anyway, to look at the method documentation...

How verbose will it be if we list out all the imports explicitly?

qctrlopencontrols/__init__.py
```diff
@@ -14,14 +14,21 @@
...
+from . import driven_controls
+from .driven_controls import *
+from . import qiskit
+from .qiskit import *
+from . import cirq
+from .cirq import *
+
+__all__ = []
```
charmasaur: Similarly, this would be more explicit if we could just list out all the appropriate methods/constants/classes.

qctrlopencontrols/driven_controls/driven_control.py
rajibchakravorty: I like that too; in that case we can also supply a list of attributes instead of dicts; we then collect both the name and the attribute values from the object (and using the list)

rajibchakravorty: Updated for review

qctrlopencontrols/driven_controls/__init__.py
rajibchakravorty: Updated; Made the list with one liberty [ALL CONSTANTS, AllClasses, all_methods] within each, alphabetical order in maintained.

rajibchakravorty: I cannot remember either; However, though it might be useful to declare __all__ for sub-packages; Can decide otherwise too;

qctrlopencontrols/__init__.py
rajibchakravorty: This is related to the other 2 comments [above and below this one]; if we decide __all__ goes only in the top level, this should take care of this problem; Otherwise, I was following numpy case e.g. https://github.com/numpy/numpy/blob/master/numpy/__init__.py; again happy to go with either decision.

charmasaur: OK, happy to be consistent with numpy. If we find it awkward going forwards we can always change it.

rajibchakravorty:
1. Made the public API (methods and classes) explicit in top-level __init__
2. Constants in sub-packages are transferred to sub-packages' own __init__

charmasaur:
Nice, looks good. The explicit relative paths used internally feel good to me -- nice and clear, and also mean that "qctrlopencontrols.<foo>" should only appear when <foo> is part of the public API, and that's true both inside and outside the package.

PR #14, opened 2019-07-02
pyQuil support in Open-Controls

rajibchakravorty:
Method to convert DDS to `pyquil.Program` added within `pyquil` sub-package.


qctrlopencontrols/pyquil/program.py
```diff
@@ -0,0 +1,209 @@
...
+        azimuthal_angle = operations[1, operation_idx]
+        x_rotation = rabi_rotation * np.cos(azimuthal_angle)
+        y_rotation = rabi_rotation * np.sin(azimuthal_angle)
+        z_rotation = operations[2, operation_idx]
+
+        rotations = np.array([x_rotation, y_rotation, z_rotation])
+        zero_pulses = np.isclose(rotations, 0.0).astype(np.int)
+        nonzero_pulse_counts = 3 - np.sum(zero_pulses)
+        if nonzero_pulse_counts > 1:
+            raise ArgumentsValueError(
```
charmasaur: Again, this isn't really an ArgumentsValueError is it? In this case I think it means *we* messed something up, because we produced an invalid DDS?

qctrlopencontrols/pyquil/program.py
```diff
@@ -0,0 +1,209 @@
...
+                program += I(qubit)
+            else:
+                if not np.isclose(rotations[0], 0.0):
+                    program += RX(rotations[0], qubit)
+                elif not np.isclose(rotations[1], 0.0):
+                    program += RY(rotations[1], qubit)
+                elif not np.isclose(rotations[2], 0.):
+                    program += RZ(rotations[2], qubit)
+
+        if np.isclose(np.sum(rotations), 0.0):
```
charmasaur: Instead could we check nonzero_pulse_counts == 0? That would be more consistent with the check above.

qctrlopencontrols/pyquil/program.py
```diff
@@ -0,0 +1,209 @@
...
+            time_covered = offsets[operation_idx]
```
charmasaur: Do we need to add gate_time here?

tests/test_pyquil_sequence.py
```diff
@@ -0,0 +1,67 @@
...
+from pyquil.quil import Pragma
+
+from qctrlopencontrols import (
+    DynamicDecouplingSequence,
+    convert_dds_to_pyquil_program)
+
+def test_pyquil_program():
+
+    """Tests if the Dynamic Decoupling Sequence gives rise to Identity
+    operation in Qiskit
```
charmasaur: Pyquil?

qctrlopencontrols/pyquil/program.py
```diff
@@ -0,0 +1,209 @@
...
+    """Converts a Dynamic Decoupling Sequence into quantum program
+    as defined in Pyquil
+
+    Parameters
+    ----------
+    dynamic_decoupling_sequence : DynamicDecouplingSequence
+        The dynamic decoupling sequence
+    target_qubits : list, optional
+        List of integers specifying target qubits for the sequence operation;
+        defaults to None
```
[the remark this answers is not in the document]
rajibchakravorty: Explained now.

qctrlopencontrols/pyquil/program.py
```diff
@@ -0,0 +1,209 @@
...
+        raise ArgumentsValueError('No dynamic decoupling sequence provided.',
+                                  {'dynamic_decoupling_sequence': dynamic_decoupling_sequence})
+
+    if not isinstance(dynamic_decoupling_sequence, DynamicDecouplingSequence):
+        raise ArgumentsValueError('Dynamical decoupling sequence is not recognized.'
+                                  'Expected DynamicDecouplingSequence instance',
+                                  {'type(dynamic_decoupling_sequence)':
+                                       type(dynamic_decoupling_sequence)})
+
+    if target_qubits is None:
```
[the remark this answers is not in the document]
rajibchakravorty: Done; This was an issue in the other conversion methods too; so changed those also.

qctrlopencontrols/pyquil/program.py
```diff
@@ -0,0 +1,209 @@
...
+        target_qubits = [0]
+
+    if gate_time <= 0:
+        raise ArgumentsValueError(
+            'Time delay of identity gate must be greater than zero.',
+            {'gate_time': gate_time})
+
+    if np.any(target_qubits) < 0:
+        raise ArgumentsValueError(
+            'Every target qubits index must be positive.',
```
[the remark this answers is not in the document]
rajibchakravorty: Corrected.

qctrlopencontrols/pyquil/program.py
```diff
@@ -0,0 +1,209 @@
...
+        offset_distance = offsets[operation_idx] - time_covered
+
+        if np.isclose(offset_distance, 0.0):
+            offset_distance = 0.0
+
+        if offset_distance < 0:
+            raise ArgumentsValueError("Offsets cannot be placed properly",
+                                      {'sequence_operations': operations})
+
+        if offset_distance > 0:
```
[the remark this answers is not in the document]
rajibchakravorty: Correct; I translated pseudocode literally. Modified; This is also an issue affecting other conversion methods; so changed those also.

qctrlopencontrols/pyquil/program.py
```diff
@@ -0,0 +1,209 @@
...
+        detuning_rotations = detuning_rotations[np.newaxis, :]
+
+    operations = np.vstack((rabi_rotations, azimuthal_angles, detuning_rotations))
+    offsets = dynamic_decoupling_sequence.offsets
+
+    time_covered = 0
+    program = Program()
+    program += Pragma('PRESERVE_BLOCK')
+
+    for operation_idx in range(operations.shape[1]):
```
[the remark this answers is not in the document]
rajibchakravorty: Corrected in all of the conversion methods;

qctrlopencontrols/pyquil/program.py
rajibchakravorty: This and the next one are related. We need to add the gate_time here because the gates will also take time to execute. What is not required is the check `np.isclose(np.sum(rotations), 0.0)` - because even if the desired unitary is Identity it should add to the time.

qctrlopencontrols/pyquil/program.py
rajibchakravorty: See above.

tests/test_pyquil_sequence.py
rajibchakravorty: Corrected;

qctrlopencontrols/pyquil/program.py
```diff
@@ -0,0 +1,209 @@
...
+
+        offset_distance = offsets[operation_idx] - time_covered
+
+        if np.isclose(offset_distance, 0.0):
+            offset_distance = 0.0
+
+        if offset_distance < 0:
+            raise ArgumentsValueError("Offsets cannot be placed properly",
```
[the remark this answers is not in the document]
rajibchakravorty: That is valid point; However, Open-Controls for now only has `ArgumentsValueError`

qctrlopencontrols/pyquil/program.py
rajibchakravorty: Again, Open-Controls only has `ArgumentsValueError`

rajibchakravorty:
Note to @michaelhush 
1) Due to comments (valid points) here changes are made in other conversion methods to make everything consistent.
2) There is a comment on ArgumentsValueError being not-suitable for some errors; I agree; Do you want to consider building some more suitable Exception/Error classes?

qctrlopencontrols/pyquil/program.py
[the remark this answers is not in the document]
charmasaur: Ah, didn't realise that. Let's wait and see what Michael says, but I don't think we need to block this PR on that (since we'll need to make changes across the board).

qctrlopencontrols/pyquil/program.py
charmasaur: Even better, nice catch.

charmasaur:
Also, to be clear, I'm not proposing we make that CircuitBuilder change now, even if it does happen to be a good idea. Just putting it out there as a possible future project.

rajibchakravorty:
> Also, to be clear, I'm not proposing we make that CircuitBuilder change now, even if it does happen to be a good idea. Just putting it out there as a possible future project.

@charmasaur - I agree. Now that we have support for 3 libraries and know what is needed - it could be done the way you are suggesting. There are subtle variations (e.g. target qubits, cirq has a Schedule etc.) that may be unique to a certain library. Therefore, even if not now, we can clean up the code a whole lot more by removing duplications where possible.

qctrlopencontrols/pyquil/program.py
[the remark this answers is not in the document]
michaelhush: I think ArgumentsValueError is right, the DDS is the problem here. But:
1) The error message is not clear enough, you should articulate specifically what the problem with the DDS is, in a manner that can help figure out how to fix it. It appears this occurs when the spacing of the gates is smaller than the gate block time? Is that right?
2) The dictionary returned with the ArgumentsValueError should always be a list of strings with the exact name of the argument that is a problem. So in this case it should at least be the dynamical_decoupling_sequence. The extras keyword can be used to provide a dictionary with other additional useful information.
3) Why are you taking a str? You should be returning the object itself in the Arguments Value Error. If it's to avoid printing the repr rather than the str, I think the repr should be printed, it's typically more informative for diagnosing errors.

qctrlopencontrols/pyquil/program.py
michaelhush: See before.

Specifically here, the offset, rabi_rotation, azimuthal_angle and detuning rotation should all be in the extras dictionary.

qctrlopencontrols/pyquil/program.py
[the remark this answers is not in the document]
charmasaur: You're right. Somehow I'd thought that we were generating the DDS internally in this method.

PR #21, opened 2019-09-02
Fix for wamf1

ViktorPerunicic:
Updated definition of wamf1.


rajibchakravorty:
@ViktorPerunicic - was the earlier implementation wrong - in the sense that the pulses did not produce required rotations etc? Or that the implementation was not in line with documentation etc.? If first, can we introduce a test for total rotation so that any change in the pulses capture the fact that the pulses can produce required rotations (or other transformations), of course limited to `pi` or `pi/2` etc. Or is there a third reason for this change?

ViktorPerunicic:
@rajibchakravorty I'd say both: it was not how the pulse is defined in docs (& in Harrison's paper), and it did not produce the correct rotations. Note, the tests were passing because the rotations it was testing for were derived from the wrong definition of the pulse.  Also, the filter function produced by the pulse didn't make sense.

PR #28, opened 2019-10-28
Lazy loading

rajibchakravorty:
Lazy loads Qiskit libraries

Lazy loads Qiskit libraries (in qiskit/quantum_circuit.py
Custom exception classes introduced (for package not found, package minimum version not found and error while importing)
NOTE: libraries from poetry and pyproject still not removed just to let CircleCI lets the test pass. Moreover, kept the required minimum version of `qiskit` a bit older for the same reason.

Comment on the pattern of the lazy loading - if it works, I will repeat this for pyquil and cirq too.


qctrlopencontrols/qiskit/quantum_circuit.py
```diff
@@ -165,8 +180,8 @@ def convert_dds_to_qiskit_quantum_circuit(
 
         if offset_distance < 0:
             raise ArgumentsValueError(
-                "Offsets cannot be placed properly. Spacing between the rotations"
-                "is smaller than the time required to perform the rotation. Provide"
+                "Offsets cannot be placed properly. Spacing between the rotations "
```
charmasaur: I'd remove all these unrelated changes from the PR -- keep it focused on the lazy loading.

charmasaur:
"Moreover, kept the required minimum version of qiskit a bit older for the same reason." -- does that mean the tests failed with a newer qiskit version? Why is that?

qctrlopencontrols/qiskit/quantum_circuit.py
rajibchakravorty: Ah did not mean to do that; reverted all these cosmetic changes.

rajibchakravorty:
> "Moreover, kept the required minimum version of qiskit a bit older for the same reason." -- does that mean the tests failed with a newer qiskit version? Why is that?

@charmasaur 

This might be related to this entry in Poetry 
```
[[package]]
category = "main"
description = "Software for developing quantum computing programs"
name = "qiskit-terra"
optional = false
python-versions = ">=3.5"
version = "0.8.2"
```
I suspect this means CircleCI installs version 0.8.2; so if we check for later version e.g. 0.10.2 when the tests are running, that check will raise the error [not the actual tests]

rajibchakravorty:
> For the lazy loading itself (i.e. making open controls faster to import) this looks like a good approach to me, but as per the Slack discussion I don't think this will help remove the actual qiskit dependency from the project. To do that, we'd need to remove it from pyproject.toml, but:
> 
> 1. if we remove qiskit from pyproject.toml, how will the tests run? CircleCI doesn't know that it needs to install qiskit.
> 2. we lose the automatic dependency solving provided by poetry/pip. For example, let's say we require a certain range of qiskit versions, which implies a certain range of jsonschema versions. And let's say a user wants to use open-controls+qiskit. The user then needs to manually figure out which range of jsonschema versions is valid with both their own requirements, and the requirements of open-controls.
> 
> I guess (2) was always going to be a drawback of this approach, so maybe we're OK with it (provided we state the allowable versions in the error message). Not sure what to do about (1) though... Maybe @stevegore has ideas. If not we might have to explore splitting up the package.

@charmasaur - Agreed. At this stage, I have started to think this needs to be resolved before we do anything.

charmasaur:
Ah,sorry, I thought you meant the minimum version of qiskit as specified in pyproject.toml (rather than in the code). Makes sense then.

stevegore:
Looks like what you're calling lazy loading, could also be called conditional importing. Am I right there?

Regardless, the whole idea behind `requirements.txt`/`setup.py`/`pyproject.toml` is that it lists _everything_ required to run a package. I don't think leaving somethings out with a note to "please install manually if you want xyz functionality" is a good pattern to follow.

I can't see any elegant way out of this at the moment. Here's a SO question saying "it can't be done" essentially: https://stackoverflow.com/questions/6570635/installing-multiple-versions-of-a-package-with-pip

I'll keep looking for a bit.

charmasaur:
Yep, I'd say conditional importing is the same as what we're trying to accomplish here.

I think it's important to be clear about our goals. There are a few problems we'd like to solve eventually:
1) `qctrl-open-controls` is slow to import, because it eagerly imports the (large) `cirq`, `qiskit` and `pyquil` libraries even though most users don't need them
2) all users of `qctrl-open-controls` need to have `cirq`, `qiskit` and `pyquil` present in their environments, even though most users don't need them
3) `qctrl-open-controls` is currently incompatible with `jobs`, due to the narrow range of `jsonschema` versions enforced by `qctrl-open-controls` via the `qiskit` dependency

Problem 1 is simple to solve, using conditional/lazy importing as implemented in this PR.

Problem 3 would disappear if we solved problem 2, but can be solved in other ways too. For example, IIUC we can simply update the `qiskit` version used by `qctrl-open-controls`, and that will open up the allowable `jsonschema` versions (see https://github.com/Qiskit/qiskit-terra/pull/3037). As a solution to the actual problem 2 at hand, therefore, I think updating `qiskit` is perfectly reasonable. Of course, further incompatibilities could be introduced in the future, so it'd be nicer if we could break the dependency of `python-core` on `jsonschema` altogether (i.e. solve problem 2), but that's true for any dependency.

Since solving 2 isn't urgent, I propose we proceed with something like this PR for speeding up the import, update the `qiskit` version to make `jobs` happy, and think about whether it makes sense to break the conversion stuff out of open controls to solve problem 2 (which isn't as urgent).


As I write this, all this has been covered in Slack, but I've invested so much time into it I'm posting it anyway :)

charmasaur:
And to expand a bit more: as it stands, I think this PR is solving problem 1 but also trying to solve 2/3 (since I think before today, I (and Rajib too maybe?) didn't realise some of the issues that would come up if we just dropped the dependencies from `requirements.txt`/`pyproject.toml`, which would have "solved" all three problems).

With that in mind, IMO we should just cut it back to basic lazy loading / conditional importing (so we don't need to worry about catching exceptions and whatnot), which will cleanly solve problem 1.

And again, y'all beat me to it on Slack.

rajibchakravorty:
Active PR coming up.

PR #29, opened 2019-10-28
Update the required version of Qiskit libraries

rajibchakravorty:
Changes to `poetry.lock` and `pyproject.toml` to ensure latest `qiskit` libraries are used.


stevegore:
Not a huge thing, but you can run `poetry update <packagename>` to only update a given dependency rather than all of them.

(tbh though I like keeping other packages updated)

PR #35, opened 2019-11-27
Added figure compatibility with qctrl-visualizer

leoadec:

- Added an export method to driven control to allow plotting with qctrl-visualizer.
- Added an export method to dynamical decoupling sequences to allow plotting with qctrl-visualizer.
- Fixed an issue with the get_plot_formatted_arrays method in driven_controls that caused the wrong graph to be plotted in cylindrical coordinates.

@qctrl/support

### DIFF
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -472,6 +472,65 @@ class DrivenControl(object):   #pylint: disable=too-few-public-methods
                                                   file_type=file_type,
                                                   coordinates=coordinates)
 
+    def export(self, coordinates=CYLINDRICAL, dimensionless_rabi_rate=True):
+        """ Returns a dictionary formatted for plotting using the qctrl-visualizer package.
+
+        Parameters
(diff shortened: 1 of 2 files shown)


qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -472,6 +472,65 @@ def export_to_file(self, filename=None,
...
+
+        plot_x = self.amplitude_x/normalizer
+        plot_y = self.amplitude_y/normalizer
+        plot_r = self.rabi_rates/normalizer
+        plot_theta = self.azimuthal_angles
+        plot_durations = self.durations
+        plot_detunings = self.detunings
+
+        if coordinates==CARTESIAN:
+            plot_dictionary["X amplitude"] = [{'value': v, 'duration': t}
```
rajibchakravorty: Here we are using `X Amplitude` as opposed to `x_amplitude`. However, in the `get_plot_formatted_arrays` it is `rabi_rates` and not `Rabi Rates`.  I would go for one policy everywhere.

PR #62, opened 2020-03-23
Miscellaneous typo corrections

leoadec:
Just a few typos I encountered during the last task.

I wonder if it might be worth running a spell checker of some kind before in a public repo like this one. Maybe it's possible to set one up like we do with pylint.


leoadec:
Yeah, we would probably need an additional quantum dictionary. An upside would be that it would be easier to keep all the names on-brand (Q-CTRL, BOULDER OPAL, etc.). I'm not sure how that feasible that is, though, but I'll ping you in case I find something interesting

PR #67, opened 2020-03-31
[QENG-714] Adding a minimum_segment_duration to DDS to controls converter

leoadec:
Partially fixes
https://q-ctrl.atlassian.net/browse/QENG-714

and was sparked by the discussion in 
https://github.com/qctrl/python-core/pull/322

I added an optional `minimum_segment_duration` to the DDS -> DrivenControls converter. The idea is: if the pulse duration estimated from the maximum frequency is smaller than the minimum segment value selected, then the `minimum_segment_duration` is used instead. The amplitudes of the controls are then calculated from the total rotation we want to achieve and the duration of the pulse, rather than always taking the maximum possible value.

I was maybe a bit conservative in these changes, and didn't touch other aspects of this function, such as the requirement that no pulse can simultaneously have Rabi and detuning rotations (alhtough I added a line in docstring to make this requirement explicit). We might want to create a ticket for a new feature supporting arbitrary rotation axes, though.

I also added three new tests in which I convert pulses with various values of the Rabi rotation, azimuthal angle, and detuning rotation. I always set the minimum_segment_duration to be the average size of those pulses, to make sure the duration of at least half of them should be affected by the new feature.

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
@@ -99,20 +99,24 @@ def convert_dds_to_driven_control(
         dynamic_decoupling_sequence=None,
         maximum_rabi_rate=2*np.pi,
         maximum_detuning_rate=2*np.pi,
+        minimum_segment_duration=0.,
         **kwargs):
     """Creates a Driven Control based on the supplied DDS and
     other relevant information
 
     Parameters
     ----------
-    dynamic_decoupling_sequence : qctrlopencontrols.DynamicDecouplingSequence
+    dynamic_decoupling_sequence : qctrlopencontrols.DynamicDecouplingSequence, optional
         The base DDS; Defaults to None
     maximum_rabi_rate : float, optional
-        Maximum Rabi Rate; Defaults to 1.0
+        Maximum Rabi Rate; Defaults to 2*pi
     maximum_detuning_rate : float, optional
-        Maximum Detuning Rate; Defaults to None
+        Maximum Detuning Rate; Defaults to 2*pi
+    minimum_segment_duration : float, optional
+        If set, further restricts the duration of every segment of the Driven Controls.
+        Defaults to 0, in which case it does not affect the duration of the pulses.
     kwargs : dict, optional
-        options to make the corresponding filter type.
+        Options to make the corresponding filter type.
         I.e. the options for primitive are described in doc for the PrimitivePulse class.
 
     Returns
@@ -132,7 +136,7 @@ def convert_dds_to_driven_control(
     -----
     Driven pulse is defined as a sequence of control segments. Each segment performs
     an operation (rotation around one or more axes). While the dynamic decoupling
-    sequence operation contains ideal instant operations, maximum rabi (detuning) rate
+    sequence operation contains ideal instant operations, the maximum Rabi (detuning) rate
     defines a minimum time required to perform a given rotation operation. Therefore, each
     operation in sequence is converted to a flat-topped control segment with a finite duration.
     Each offset is taken as the mid-point of the control segment and the width of the
@@ -141,7 +145,7 @@ def convert_dds_to_driven_control(
     If the sequence contains operations at either of the extreme ends
     :math:`\\tau_0=0` and :math:`\\tau_{n+1}=\\tau`(duration of the sequence), there
     will be segments outside the boundary (segments starting before :math:`t<0`
-    or finishing after the sequence duration :math:`t>\\tau`. In these cases, the segments
+    or finishing after the sequence duration :math:`t>\\tau`). In these cases, the segments
     on either of the extreme ends are shifted appropriately so that their start/end time
     falls entirely within the duration of the sequence.
 
@@ -221,20 +225,17 @@ def convert_dds_to_driven_control(
             continue
 
         if operations[3, op_idx] == 0: #no z_rotations
-            if not np.isclose(operations[1, op_idx], 0.):
-                half_pulse_duration = 0.5 * operations[1, op_idx] / maximum_rabi_rate
-            else:
-                half_pulse_duration = 0.5 * operations[2, op_idx] / maximum_rabi_rate
+            half_pulse_duration = 0.5 * max(operations[1, op_idx] / maximum_rabi_rate,
+                                            minimum_segment_duration)
 
             pulse_start_ends[op_idx, 0] = pulse_mid_points[op_idx] - half_pulse_duration
-
             pulse_start_ends[op_idx, 1] = pulse_mid_points[op_idx] + half_pulse_duration
         else:
-            pulse_start_ends[op_idx, 0] = pulse_mid_points[op_idx] - \
-                                          0.5 * operations[3, op_idx] / maximum_detuning_rate
+            half_pulse_duration = 0.5 * max(operations[3, op_idx] / maximum_detuning_rate,
+                                            minimum_segment_duration)
 
-            pulse_start_ends[op_idx, 1] = pulse_mid_points[op_idx] + \
-                                          0.5 * operations[3, op_idx] / maximum_detuning_rate
+            pulse_start_ends[op_idx, 0] = pulse_mid_points[op_idx] - half_pulse_duration
+            pulse_start_ends[op_idx, 1] = pulse_mid_points[op_idx] + half_pulse_duration
 
     # check if any of the pulses have gone outside the time limit [0, sequence_duration]
     # if yes, adjust the segment timing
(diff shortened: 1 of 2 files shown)


tests/test_dynamical_decoupling.py
```diff
@@ -437,6 +437,193 @@ def test_conversion_of_z_pi_2_pulses_at_extremities():
...
 
+
+def test_conversion_of_pulses_with_arbitrary_rabi_rotations():
+    """
+    Tests if the method to convert a DDS to driven controls handles properly
+    Y pulses with rabi rotations that assume arbitrary values between 0 and pi.
+    """
+    _number_of_pulses = 30
+    _duration = _number_of_pulses * 1.
+    _offsets = np.linspace(0.5, _duration-0.5, _number_of_pulses)
```
charmasaur: When comparing to the expected durations defined below, it took a little while to convince myself about what the spacing is here. I think I would have found it simpler to understand if this was just `0.5 + np.arange(_number_of_pulses)`.

tests/test_dynamical_decoupling.py
```diff
@@ -437,6 +437,193 @@ def test_conversion_of_z_pi_2_pulses_at_extremities():
         [2.5e-2, _duration - 2* 2.5e-2, 2.5e-2]))
 
 
+
+def test_conversion_of_pulses_with_arbitrary_rabi_rotations():
```
charmasaur: Just a question/comment, no action required (unless you want): do you think we get that much value out of having such a large number of pulses here? The advantage to using a smaller number, like 2 or 3, is that you can write everything out explicitly, which can make it quite a bit simpler to follow the test (and also means you're less likely to get something wrong in the test).

tests/test_dynamical_decoupling.py
```diff
@@ -437,6 +437,193 @@ def test_conversion_of_z_pi_2_pulses_at_extremities():
...
+
+
+def test_conversion_of_pulses_with_arbitrary_azimuthal_angles():
+    """
+    Tests if the method to convert a DDS to driven controls handles properly
+    pi-pulses with azimuthal angles that assume arbitrary values between 0 and pi/2.
+    """
+    _number_of_pulses = 30
+    _duration = _number_of_pulses * 1.
+    _offsets = np.linspace(0.5, _duration-0.5, _number_of_pulses)
```
charmasaur: Same comment as above (and same below too)

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -99,20 +99,27 @@ def convert_dds_to_driven_control(
...
+    """
+    Creates a Driven Control based on the supplied DDS and other relevant information.
+
+    Currently, pulses that simultaneously contain Rabi and detuning rotations are not
+    supported.
 
     Parameters
     ----------
-    dynamic_decoupling_sequence : qctrlopencontrols.DynamicDecouplingSequence
+    dynamic_decoupling_sequence : qctrlopencontrols.DynamicDecouplingSequence, optional
```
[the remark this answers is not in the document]
leoadec: Oops, my bad! I've now reverted that change. I think your suggestion of removing the "Defaults to None" also makes sense, so I've done it as well. As for the third point, I think it's reasonable, but I'm not 100% sure either if nothing will change if we change the error that is returned when the user doesn't provide a DDS (it's *probably* fine though, so maybe I'm 90% sure it is okay). Still, maybe we can leave this third point as is for the moment

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -216,22 +223,18 @@ def convert_dds_to_driven_control(
...
+        if not np.isclose(operations[1, op_idx], 0.): # Rabi rotation
+            half_pulse_duration = 0.5 * max(operations[1, op_idx] / maximum_rabi_rate,
+                                            minimum_segment_duration)
+        elif not np.isclose(operations[3, op_idx], 0.): # Detuning rotation
+            half_pulse_duration = 0.5 * max(np.abs(operations[3, op_idx]) / maximum_detuning_rate,
+                                            minimum_segment_duration)
 
-            pulse_start_ends[op_idx, 1] = pulse_mid_points[op_idx] + half_pulse_duration
+        pulse_start_ends[op_idx, 0] = pulse_mid_points[op_idx] - half_pulse_duration
+        pulse_start_ends[op_idx, 1] = pulse_mid_points[op_idx] + half_pulse_duration
```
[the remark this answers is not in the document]
leoadec: Yes, it may, and yes, this is one of the reason for the extra test below. The other reason is that there's no guarantee that the spacings between the pulses will be sufficient to satisfy a minimum_segment_duration even before that shift

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -270,6 +273,18 @@ def convert_dds_to_driven_control(
                                   extras={'deduced_pulse_start_timing': pulse_start_ends[:, 0],
                                           'deduced_pulse_end_timing': pulse_start_ends[:, 1]})
 
+    # check if the minimum_segment_duration is respected in the gaps between the pulses
+    if (np.any(pulse_start_ends[1:, 0]-pulse_start_ends[:-1, 1] < minimum_segment_duration)):
```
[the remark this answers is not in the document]
leoadec: I think that's a fair point, I've re-written the error to allow some tolerance for the floating point imprecisions. (It would be nice if we could have `less_or_close` and `greater_or_close` as functions we can import from somewhere so we don't have to effectively re-write them in every file.)

I think you're right that the lack of floating point tolerance is what made Luigi's test fail yesterday. I think it might also be worth rewriting the tests above so that the error messages are a little bit more helpful as well (at the moment they don't say which of the four conditions was violated). I feel this is a bit beyond the scope of this PR, so maybe I can do it next and add it to the same release of Open Controls as this

tests/test_dynamical_decoupling.py
leoadec: Oh, true. I basically only used that large number because I wanted to convince myself that things were working properly. I've now reduced the number of pulses to three in each sequence. In the Rabi/detuning tests I tried to make it so one of the pulses was limited by the upper_limit of the controls, one was limited by the minimum_segment_duration, and one had these two values in perfect balance. Let me know if it's more readable now

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -270,6 +273,20 @@ def convert_dds_to_driven_control(
                                   extras={'deduced_pulse_start_timing': pulse_start_ends[:, 0],
                                           'deduced_pulse_end_timing': pulse_start_ends[:, 1]})
 
+    # check if the minimum_segment_duration is respected in the gaps between the pulses
+    if not np.any(np.logical_or(np.greater(pulse_start_ends[1:, 0]-pulse_start_ends[:-1, 1],
+                                           minimum_segment_duration),
+                                np.isclose(pulse_start_ends[1:, 0], pulse_start_ends[:-1, 1]))):
```
charmasaur, line 279: Shouldn't this be checking `np.isclose(pulse_start_ends[1:, 0]-pulse_start_ends[:-1, 1], minimum_segment_duration)`?

May be a little more readable if before you start this `if` check you make a local var for the gap durations:
`gap_durations = pulse_start_ends[1:, 0]-pulse_start_ends[:-1, 1]`

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -270,6 +273,20 @@ def convert_dds_to_driven_control(
                                   extras={'deduced_pulse_start_timing': pulse_start_ends[:, 0],
                                           'deduced_pulse_end_timing': pulse_start_ends[:, 1]})
 
+    # check if the minimum_segment_duration is respected in the gaps between the pulses
+    if not np.any(np.logical_or(np.greater(pulse_start_ends[1:, 0]-pulse_start_ends[:-1, 1],
```
charmasaur, line 277: Hm, I'm not sure this is right. The "or" is checking "is the gap >= minimum_segment_duration". The "any" is checking "are any of the gaps >= minimum_segment_duration". I think we want "all" instead?

tests/test_dynamical_decoupling.py
charmasaur: Nice! Yeah, I find these easier to follow. And it's also more clear that you're specifically triggering all the cases.

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
leoadec: Ooops, sorry, I really messed this up. Yeah, should be `np.all` and `np.isclose` should be comparing the gap with `minimum_segment_duration` rather than with each other (I was thinking about Luigi's problem when I was writing this and I guess I got crossed wires in my brain haha). I adopted your `gap_durations` array to make things less confusing to myself, thanks

PR #69, opened 2020-04-01
Making the errors in DDS to Controls conversion more floating-point-friendly

leoadec:
Final part of the fix for the bug found by Luigi. First part was PR #68

I split the checks of the timing of the pulses into two parts:
1. Checking that all the offsets are in ascending order in time (and adding the information about this requirement to the docstring).
2. Checking that none of the pulses overlap.

These two checks should be tolerant of floating point imprecisions now.

I added a unit test that replicates exactly the situation that Luigi reported. In the master branch, this test fails with the exception:

```
ArgumentsValueError: Pulse timing could not be properly deduced from the sequence operation offsets. Try increasing the maximum rabi rate or maximum detuning rate.
```

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
@@ -238,14 +238,11 @@ def convert_dds_to_driven_control(
         translation = pulse_start_ends[-1, 1] - sequence_duration
         pulse_start_ends[-1, :] = pulse_start_ends[-1, :] - translation
 
-    # four conditions to check
+    # two conditions to check
     # 1. Control segment start times should be monotonically increasing
     # 2. Control segment end times should be monotonically increasing
-    # 3. Adjacent segments should not be overlapping
     if (np.any(pulse_start_ends[0:-1, 0] - pulse_start_ends[1:, 0] > 0.) or
-            np.any(pulse_start_ends[0:-1, 1] - pulse_start_ends[1:, 1] > 0.) or
-            np.any(pulse_start_ends[1:, 0]-pulse_start_ends[0:-1, 1] < 0.)):
-
+        np.any(pulse_start_ends[0:-1, 1] - pulse_start_ends[1:, 1] > 0.)
         raise ArgumentsValueError('Pulse timing could not be properly deduced from '
                                   'the sequence operation offsets. Try increasing the '
                                   'maximum rabi rate or maximum detuning rate.',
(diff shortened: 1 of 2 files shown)


qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -206,6 +206,16 @@ def convert_dds_to_driven_control(
         azimuthal_angles = np.append(azimuthal_angles, [0])
         detuning_rotations = np.append(detuning_rotations, [0])
 
+    # check that the offsets are correctly sorted in time
+    time_differences = np.diff(offsets)
+    if not np.all(np.logical_or(np.greater(time_differences, 0.),
```
charmasaur, line 211: In this case I'm not convinced that we should allow "close" offsets. Typically we'd use a "close" check when equality is allowed, but here I think we want the offsets to be strictly increasing rather than non-decreasing.

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -206,6 +206,16 @@ def convert_dds_to_driven_control(
...
+                                np.isclose(time_differences, 0.))):
+        raise ArgumentsValueError("Pulse timing could not be properly deduced from "
+                                  "the sequence offsets. Make sure all offset are "
+                                  "correctly ordered in time.",
```
charmasaur, line 215: Instead of saying "correctly ordered in time", how about we just say "increasing order"? That would be a more clear indication of how the user can fix the issue.

tests/test_dynamical_decoupling.py
```diff
@@ -575,6 +575,42 @@ def test_conversion_of_pulses_with_arbitrary_detuning_rotations():
     assert _all_greater_or_close(driven_control.duration, minimum_segment_duration)
 
 
+def test_conversion_of_tightly_packed_sequence():
+    """
+    Tests if the method to convert a DDS to driven controls handles properly
+    a sequence tightly packed with pulses, where there is no time for a gap between
+    the pi/2-pulses and the adjacent pi-pulses.
+    """
+    # create a sequence containing 30 pi-pulses and 2 pi/2-pulses at the extremities
```
charmasaur, line 584: Similar to the comment on the last PR, but I think we'd get the same amount of coverage with a sequence of maybe 4 offsets (pi/2, pi, pi, pi/2). In the context of the bug Luigi hit it's nice to verify that situation exactly, but for the test that we actually merge I'd suggest favouring simplicity.

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -244,25 +254,20 @@ def convert_dds_to_driven_control(
...
-                                   'maximum_rabi_rate': maximum_rabi_rate,
-                                   'maximum_detuning_rate': maximum_detuning_rate},
-                                  extras={'deduced_pulse_start_timing': pulse_start_ends[:, 0],
-                                          'deduced_pulse_end_timing': pulse_start_ends[:, 1]})
+    # check that no adjacent pulses overlap
+    gap_durations = pulse_start_ends[1:, 0] - pulse_start_ends[:-1, 1]
+    if not np.all(np.logical_or(np.greater(gap_durations, 0.),
+                                np.isclose(gap_durations, 0.))):
+            raise ArgumentsValueError("There is overlap between pulses in the sequence. "
+                                      "Try increasing the maximum rabi rate or maximum detuning rate.",
```
charmasaur, line 262: Is this necessarily the issue? I think it could also be `minimum_segment_duration`, if that's the limiting factor for the pulse durations.

If that's right, an alternative would be to drop this check, and update the error message below to also suggest increasing the rabi/detuning rate.

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -119,6 +119,7 @@ def convert_dds_to_driven_control(
     minimum_segment_duration : float, optional
         If set, further restricts the duration of every segment of the Driven Controls.
         Defaults to 0, in which case it does not affect the duration of the pulses.
+        Must be greater or equal to 0, if set.
```
charmasaur, line 122: Good point! I think this is a reasonably requirement.

leoadec: Yeah, I was getting away without testing it explicitly before, but if I eliminate the gap > 0 test then I have to make sure this isn't negative

leoadec:
@charmasaur Thanks for the feedback and for the fast approval! I guess you were convinced that this was correct faster than I was haha

PR #72, opened 2020-04-08
Fix lint issues

tachikoma-li:
- Fix lint issues
- Apply BLACK to most files of the repo
- Add BLACK to poetry dependency


.github/workflows/on-push.yml
```diff
@@ -33,9 +33,9 @@ jobs:
     - name: Run Pylint
       run: |
         ./ci docker run qctrl/python-build:3.7 poetry run pylint_runner || true
```
leoadec, line 35: I think you should be able to remove the `|| true` from here once all the linting issues are solved

pyproject.toml
```diff
@@ -40,14 +40,25 @@ sphinx-rtd-theme = "^0.4.3"
...
+     /(
+         docs
+     )/
+     | setup.py
+)
+'''
+
 [tool.dephell.main]
-from = {format = "poetry", path = "pyproject.toml"}
+from ={format="poetry", path="pyproject.toml"}
```
leoadec, line 60: I'm a bit puzzled about why the formatting in this line changes, but the formatting in the `to =` line below remained the same

.github/workflows/on-push.yml
```diff
@@ -33,9 +33,9 @@ jobs:
...
-    - name: Run Pylama
+    - name: Run Black
       run: |
-        ./ci docker run qctrl/python-build:3.7 poetry run pylama || true
+        ./ci docker run qctrl/python-build:3.7 poetry run black -- --check .
```
charmasaur, line 38: nit: I'd suggest putting the Black check before pylint, because Black is a lot faster, so if there are basic issues then by doing Black first we'll be able to give a result faster

qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py
```diff
@@ -212,45 +229,50 @@ def convert_dds_to_driven_control(
...
+    )
 
     pulse_mid_points = operations[0, :]
 
-    pulse_start_ends = np.zeros((
-        operations.shape[1], 2))   # pylint: disable=unsubscriptable-object
-
-    for op_idx in range(operations.shape[1]):   # pylint: disable=unsubscriptable-object
+    pulse_start_ends = np.zeros(
+        (operations.shape[1], 2)  # pylint: disable=unsubscriptable-object
```
charmasaur, line 253: I think you can remove these, since the check is disabled in pylintrc (same with more of the same below).

pyproject.toml
tachikoma-li: I set `isort` to auto run after save. For some reason, it insists woking on `toml` file :(

docs/conf.py
```diff
@@ -13,25 +15,25 @@
...
 
 # -- HTML Variables ----------------------------------------------------------
 # Variables to insert into _templates/layout.html
 # Taken from qctrl/docs, see head.html in the repo
 html_context = {
     "var_url": "https://docs.q-ctrl.com",
     "var_title": "Open Controls Python package | Q-CTRL",
-    "var_description": "Module, class and method reference for the Q-CTRL Open Controls Python package",
+    "var_description": "Module, class and method reference forthe Q-CTRL Open Controls\
+        Python package",
```
charmasaur, line 36: If I understand Python line continuation correctly then this will introduce a bunch of spaces in the string. Instead you can use implicit string concatenation (or explicit if you prefer):
```
"var_description": ("Module, class and method reference for the Q-CTRL Open Controls "
                    "Python package")
```

Also add space between "for" and "the" :)

tachikoma-li: Nice! I didn't know the implicit string concatenation before. Thanks.

PR #74, opened 2020-04-24
Add DocsSearch-content class

stevegore:


stevegore:
Block modified from https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/layout.html#L185

PR #81, opened 2020-06-09
Fixing bug in DrivenControl.directions method

leoadec:
Currently, running

```
from qctrlopencontrols import DrivenControl
control = DrivenControl(rabi_rates=[1, 0, 1], detunings=[0, 0, 0], durations=[1, 1, 1])
control.directions
```

will result in an error, because the segment with zero amplitude (the second segment) will cause a division by zero in the `directions` method. This patch prevents that division by zero.

Let me know if you'd prefer a different solution

### DIFF
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -280,7 +280,7 @@ class DrivenControl:
         )
 
         # Reduces tolerance of the comparison to zero in case the units chosen
-        # make the amplitudes very small, but never allows it be higher than the
+        # make the amplitudes very small, but never allows it to be higher than the
         # default atol value of 1e-8
         tolerance = min(1e-20 * np.max(amplitudes), 1e-8)
 
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -278,23 +278,19 @@ class DrivenControl:
         amplitudes = np.sqrt(
             self.amplitude_x ** 2 + self.amplitude_y ** 2 + self.detunings ** 2
         )
+
+        amplitudes = np.where(np.isclose(amplitudes, 0.0), 1.0, amplitudes)
+
         normalized_amplitude_x = self.amplitude_x / amplitudes
         normalized_amplitude_y = self.amplitude_y / amplitudes
         normalized_detunings = self.detunings / amplitudes
 
-        normalized_amplitudes = np.hstack(
+        directions = np.hstack(
             (
                 normalized_amplitude_x[:, np.newaxis],
                 normalized_amplitude_y[:, np.newaxis],
                 normalized_detunings[:, np.newaxis],
             )
-        )
-
-        directions = np.array(
-            [
-                normalized_amplitudes if amplitudes[i] != 0.0 else np.zeros([3,])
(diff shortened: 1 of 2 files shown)


qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -278,25 +278,21 @@ def directions(self):
         amplitudes = np.sqrt(
             self.amplitude_x ** 2 + self.amplitude_y ** 2 + self.detunings ** 2
         )
+
+        amplitudes = np.where(np.isclose(amplitudes, 0.0), 1.0, amplitudes)
```
charmasaur, line 282: Is there a test for this class? If so, might be good just to add a small test for this case, to prevent it breaking again in the future.

abenseny, line 282: I don't know how relevant this would be here, but if all amplitudes are below 1e-8 (the default `atol` in `np.isclose`) because of the units the user is using, they'll all be flagged as close to 0.

A workaround would be to compare with "normalised amplitudes":
`np.where(np.isclose(amplitudes/np.max(amplitudes), 0.0), 1.0, amplitudes)`
but that would also fail if they're all zero.

If it's a "hard" division by zero you could also check 
`np.where(amplitudes == 0.0, 1.0, amplitudes)`?

leoadec: Sure. There was already one test for the class, but it seemed kind of bloated already, so I added another short test for the directions method

leoadec: That's a good point. I was a bit afraid of comparing to zero directly with `== 0.0` (it might not be a problem though, as probably np.sqrt rounds the amplitude down to zero when the numbers get really small, but I'm not completely confident of that), so instead I changed `np.isclose` to use a reduced `atol` that is 1e-20 of the maximum value of the amplitudes.

I'm not convinced that this is the best solution, but I probably won't manage to come up with better ideas today. For what's worth, it seems to work with the extra test I added (which uses ridiculously small amplitudes of the order of 1e-100). Other suggestions are welcome, in any case

abenseny: I like your solution, it's also probably smarter to change the `atol` than normalize the inputs. (I also can't come up with any better alternatives).

charmasaur: Note that `durations` are specified to be in seconds, so the user doesn't really have any choice for units (the amplitudes should all be in Hz).

Still, I don't know what the best thing to do is in situations like this. If we use a hard `==0.` check then we'll silently get floating point errors (and possibly `nan`s) when the user gets too close to `0` (which can also happen with the current solution, since the `1e-20` multiplication can send the tolerance down to `0`). If we use a check with a magic number then we'll silently provide an inaccurate value if the user goes too small (but that inaccurate value will still satisfy some nice properties, such as giving a unit vector).

Given all that, I'm fine with this as-is, but I'm also keen to work out some pattern we can use for this sort of thing when we inevitably hit it again.

leoadec: Yeah, I'd be keen to workshop some solution for this kind of situation, as we seem to keep running into that. I imagine other people also have run into this and come up with their own ingenious solutions, and maybe we can learn from them if we figure out where to find them. Unfortunately I ran out of ideas for today

PR #88, opened 2020-07-06
Apply isort

tachikoma-li:
use `isort` to sort import


pyproject.toml
```diff
@@ -58,6 +59,13 @@ exclude = '''
 )
 '''
 
+[tool.isort]
```
charmasaur, line 62: Is this the same config as we have internally?

tachikoma-li: yep.

PR #94, opened 2020-07-13
Fix pre-post rotations

tachikoma-li:
As discussed in https://q-ctrl.atlassian.net/browse/QENG-974
we now add x rotations unconditionally to all DDS, but the sign of the post-rotation could be different depending on the control scheme. The net effect of DDS now does not have to be identity, this is captured in tests.

I have a strange test error. All tests pass on local machines (tested on Mac and Linux). However, there seem random failures in CI. The issue is related to the function `test_if_quadratic_sequence_with_odd_inner_pulses_is_identity` in ` tests/test_predefined_dynamical_decoupling.py`. See commit `75e3a6c` for example. I notice sometimes rerun the action would resolve the issue, but I don't have a clue why it happens. @leoadec @charmasaur

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -42,7 +42,7 @@ def _add_pre_post_rotations(
     """Adds a pre and post X rotation at the start and end of the sequence.
 
     Note that with these two pre and post X rotations, the net effect of the DDS does not
-    necessarily have to be an identity. For example, given a CPMG sequences of odd number Y pi
+    necessarily have to be an identity. For example, given a CPMG sequence of odd number Y pi
     rotations in the middle with the pre (pi/2) and post(-pi/2) X rotations, the net effect will
     be a Z gate.
 
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -127,9 +127,10 @@ def _add_pre_post_rotations(
     # of Y and Z pi-pulses
     preserves_11 = (y_pi_pulses + z_pi_pulses) % 2 == 0
 
-    # use pi/2 and -pi/2 (pi/2) X pulses as pre and post rotations for CP sequences with
-    # odd(even) number of pulses
-    # always use pi/2 and -pi/2 X pulses as pre and post rotations for CPMG sequences
+    # the direction of final rotation depends on the property of DDS.
+    # if the net effect of the sequences is an identity gate or Y rotation, the post rotation
+    # is chosen to be -pi/2 X pulse, otherwise use pi/2 X pulse, to ensure the final gate is an
+    # identity or Z rotation.
     if (preserves_10 and preserves_11) or (not preserves_10 and not preserves_11):
         final_azimuthal = np.pi
 
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -127,9 +127,9 @@ def _add_pre_post_rotations(
     # of Y and Z pi-pulses
     preserves_11 = (y_pi_pulses + z_pi_pulses) % 2 == 0
 
-    # the direction of final rotation depends on the property of DDS.
+    # the direction of the post rotation depends on the property of DDS.
     # if the net effect of the sequences is an identity gate or Y rotation, the post rotation
-    # is chosen to be -pi/2 X pulse, otherwise use pi/2 X pulse, to ensure the final gate is an
+    # is chosen to be -pi/2 X pulse, otherwise use pi/2 X pulse, to ensure the net effect is an
     # identity or Z rotation.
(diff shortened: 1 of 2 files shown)


qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -126,40 +127,11 @@ def _add_pre_post_rotations(
...
 
-    # When states |0> and |0>+|1> are preserved, the sequence already produces
-    # an identity, so that we want the the pi/2-pulses to cancel each other out
-    if preserves_10 and preserves_11:
-        rabi_value = np.pi / 2
-        initial_azimuthal = 0
+    # use pi/2 and -pi/2 (pi/2) X pulses as pre and post rotations for CP sequences with
+    # odd(even) number of pulses
+    # always use pi/2 and -pi/2 X pulses as pre and post rotations for CPMG sequences
+    if preserves_10 or (not preserves_11):
```
charmasaur, line 133: Hm, I'm not sure about this. What if we had `preserves_10=True` and `preserves_11=False` (i.e. a Z rotation)? In that case, when we had the pre (+pi/2) and post (-pi/2) rotations, we'd end up with an overall X.

Instead I think we want:
10 + 11 -> -pi/2 (overall I)
10 + ~11 -> +pi/2 (overall Z)
~10 + 11 -> +pi/2 (overall I)
~10 + ~11 -> -pi/2 (overall Z)

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -39,17 +39,12 @@
...
-    The parameters of the pi/2-pulses are chosen in order to cancel out the
-    product of the pulses in the DSS, so that its total effect in the
-    absence of noise is an identity.
-
-    For a DSS that already produces an identity, this function adds X pi/2-pulses
-    in opposite directions, so that they cancel out. If the DDS produces an X
-    gate, the X pi/2-pulses will be in the same direction. If the DDS produces
-    a Y (Z) gate, the pi/2-pulses are around the Y (Z) axis.
+    Note that with these two pre and post X rotations, the net effect of the DDS does not
+    necessarily have to be an identity. For example, given a CPMG sequence of odd number Y pi
```
charmasaur, line 45: I think it'd be worth clarifying that the net effect does not have to be an identity, but it *will* always be either an identity or Z pi rotation.

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
leoadec: Maybe we can use as the condition `(preserves_10 and preserves_11) or (not preserves_10 and not preserves_11)`

tachikoma-li: ah, right, that is why I had a weird `extra_operation = \sigma_x * (-\sigma_z)` in the test. Thanks.

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -126,40 +127,11 @@ def _add_pre_post_rotations(
     # of Y and Z pi-pulses
     preserves_11 = (y_pi_pulses + z_pi_pulses) % 2 == 0
 
-    # When states |0> and |0>+|1> are preserved, the sequence already produces
-    # an identity, so that we want the the pi/2-pulses to cancel each other out
-    if preserves_10 and preserves_11:
-        rabi_value = np.pi / 2
-        initial_azimuthal = 0
+    # use pi/2 and -pi/2 (pi/2) X pulses as pre and post rotations for CP sequences with
```
charmasaur, line 130: I'd suggest updating this comment to be more general, since it's meant to work for more than just CP/CPMG sequences. IMO what's important is that for an identity or Y rotation we use a -pi/2 final rotation, to ensure that the total rotation is either I or Z.

tests/test_predefined_dynamical_decoupling.py
```diff
@@ -670,18 +674,19 @@ def test_attribute_values():
         )
 
 
-def _pulses_produce_identity(sequence):
+def _pulses_produce_identity(sequence, extra_rotation=None):
```
charmasaur, line 677: I'd tentatively suggest changing this so that it always checks for I or Z, since I think that's really the property that we're always trying to check with this function.

If we *do* want to be more prescriptive (by checking for a specific operation), I'd suggest changing this either to accept the full expected operator all the time (e.g. `def _check_pulses(sequence, expected_operator)` or something), or to return the operator produced by the pulses so that the caller can check it themselves (e.g. `def _get_operator(sequence)`).

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
tachikoma-li: done.

tests/test_predefined_dynamical_decoupling.py
tachikoma-li: SG. Remove `extra_rotation` and check explicitly Z rotation in return.

tachikoma-li:
No idea what happened, but the issue with the test seems gone with the changes in the implementation.

Issue #90, opened 2020-07-09
Disallow negative Rabi rates in DynamicDecouplingSequence


charmasaur:
At the moment `DynamicDecouplingSequence` accepts negative values for `rabi_rotations`, which is both unnecessary (the same effect could be achieved by adding `pi` to the corresponding `azimuthal_angles` values) and can cause issues later when we try to create a `DrivenControl` (see https://github.com/qctrl/python-open-controls/blob/master/qctrlopencontrols/dynamic_decoupling_sequences/driven_controls.py#L197).

I think it makes sense to disallow negative Rabi rates everywhere rather than allow them everywhere, because then there's less ambiguity about how a certain pulse should be represented.

This task is to update the `DynamicalDecouplingSequence` class to explicitly disallow negative Rabi rates, and to make sure that we're never trying to set negative Rabi rates when creating the predefined sequences here: https://github.com/qctrl/python-open-controls/blob/master/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py.

As part of this, it might make sense to introduce a shared set of validation functions (e.g. in `utils.py`) to check that a given set of Rabi rates, azimuthal angles and detunings are all valid. Those validations can then be used everywhere.

tachikoma-li:
Update: There is now a function called `check_arguments`  in `utils.py` for this task.

PR #96, opened 2020-07-15
Add a function to check arguments

tachikoma-li:
While adding `mypy` to check type, it seems some refactor for argument checking is necessary besides just adding type hints. It will be easier to add this check function first.
 
We might apply this function for all arguments check in the future. 

This is also part of the issue https://github.com/qctrl/python-open-controls/issues/90#issue-653672409

### DIFF
--- a/qctrlopencontrols/utils.py
+++ b/qctrlopencontrols/utils.py
@@ -87,13 +87,25 @@ def check_arguments(condition, description, arguments, extras=None):
     Raises an ArgumentsValueError with the specified parameters if the given condition is false,
     otherwise does nothing.
 
+    For example,  a use case may look like::
+
+        def log(x):
+            check_arguments(x > 0,
+                            "x must be positive.",
+                            {"x": x})
+            return numpy.log(x)
+
     Parameters
     ----------
-    conditions: Any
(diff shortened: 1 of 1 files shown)


qctrlopencontrols/utils.py
```diff
@@ -82,6 +82,28 @@ def create_repr_from_attributes(class_instance=None, attributes=None):
...
 
 
+def check_arguments(condition, description, arguments, extras=None):
+    """
+    Raises an ArgumentsValueError with the specified parameters if the given condition is false,
+    otherwise does nothing.
+
+    Parameters
+    ----------
+    conditions: Any
```
charmasaur, line 92: condition*

charmasaur, line 92: Add descriptions for these parameters

qctrlopencontrols/utils.py
```diff
@@ -82,6 +82,28 @@ def create_repr_from_attributes(class_instance=None, attributes=None):
...
+    description: str
+    arguments: dict
+    extra: dict, optional
```
charmasaur, line 95: extras

PR #100, opened 2020-07-24
Remove `driven_control` from dynamical decoupling module

tachikoma-li:
This is open to discussion.  Main reasons for this PR:
1. it is confusing to have a `driven_control` file inside the dynamical decoupling module while we already have a 'driven_control` module
2. this results in some confusion in imports
3. the only thing been done in the `driven_control' file is to convert dynamical sequences to driven control. It might be better to just have this function in the dynamical decoupling module itself


qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
```diff
@@ -324,3 +328,337 @@ def export_to_file(
             file_type=file_type,
             coordinates=coordinates,
         )
+
+
+def convert_dds_to_driven_control(
```
charmasaur, line 333: Is this a direct copy from the old module?

tachikoma-li: yep. Should have mentioned it in the description.

PR #102, opened 2020-07-27
Add type hint for predefined driven control

tachikoma-li:
Private methods like `_new_***` allow `rabi_rotation` to be optional but they all call `_predefined_common_attributes` which does not allow `rabi_rotation` to be `None`.
Some refactoring is done to make sure the code pass type checking.

Edit: note that this is only to fix the type issue, but does not fix the bug when user doesn't explicitly set `rabi_rotation`.

### DIFF
--- a/qctrlopencontrols/driven_controls/predefined.py
+++ b/qctrlopencontrols/driven_controls/predefined.py
@@ -2,7 +2,9 @@
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at # #     http://www.apache.org/licenses/LICENSE-2.0
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
@@ -192,7 +194,7 @@ def _derive_segments(
 ) -> List[List[float]]:
     """
     Derive the driven control segments from a set of rabi_rotations defined in terms of the
-        spherical polar angles
+    spherical polar angles.
 
     Parameters
     ----------
@@ -322,7 +324,7 @@ def _new_solovay_kitaev_1_control(
     **kwargs
 ) -> DrivenControl:
     """
-    First-order Solovay-Kitaev control, also known as SK1
+    First-order Solovay-Kitaev control, also known as SK1.
 
     Parameters
     ----------
--- a/qctrlopencontrols/driven_controls/predefined.py
+++ b/qctrlopencontrols/driven_controls/predefined.py
@@ -121,7 +121,7 @@ def _predefined_common_attributes(
         The azimuthal position of the pulse.
     rabi_rotation : float
         The total polar angle to be performed by the pulse.
-        Defined in polar coordinates. Defaults to None.
+        Defined in polar coordinates.
     maximum_rabi_rate : float, optional
         Defaults to 2.*np.pi
         The maximum rabi frequency for the pulse.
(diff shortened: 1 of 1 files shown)


qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -266,13 +289,12 @@ def _new_wimperis_1_control(
...
-    (maximum_rabi_rate, rabi_rotation, azimuthal_angle) = _predefined_common_attributes(
-        maximum_rabi_rate, rabi_rotation, azimuthal_angle
+    (maximum_rabi_rate, azimuthal_angle, rabi_rotation) = _predefined_common_attributes(
+        maximum_rabi_rate, azimuthal_angle, rabi_rotation
     )
-
-    phi_p = _get_transformed_rabi_rotation_wimperis(rabi_rotation)
-
-    rabi_rotations = [rabi_rotation, np.pi, 2 * np.pi, np.pi]
+    if rabi_rotation is not None:
```
charmasaur, line 295: Any idea why the type checker isn't picking up that line 302 could fail if `rabi_rotation` is `None`, since `phi_p` won't get defined?

tachikoma-li: Not sure. I was quite surprised to see such an obvious error just passes mypy checking.

tachikoma-li:
> Does it even make sense for the parameter to be optional? I would have thought we should just make it required (or, at the very least, make it optional with an explicit default). That would probably make the type checking a bit easier.

Seems we have an agreement here, I'll just make it required. Making `None` as default value does not make sense to me.

qctrlopencontrols/driven_controls/predefined.py
charmasaur: https://github.com/python/mypy/issues/2400

Crazy stuff.

tachikoma-li: indeed....

tachikoma-li:
Make `rabi_rotation` required and set the default value to `maximum_rabi_rate` as stated in the docstring.

qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -374,38 +402,39 @@ def _short_composite_rotation_for_undoing_length_over_under_shoot_control(
...
-        phi_1 = np.arccos(
-            -np.pi * np.cos(theta_1) / 2 / theta_1 / np.sin(rabi_rotation / 2)
-        )
-        phi_2 = phi_1 - np.arccos(-np.pi / 2 / theta_1)
-    else:
-        raise ArgumentsValueError(
-            "rabi_rotation angle must be either pi, pi/2 or pi/4",
-            {"rabi_rotation": rabi_rotation},
-        )
+    if rabi_rotation is not None:
```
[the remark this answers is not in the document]
charmasaur: Sorry, I missed this comment

qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -104,26 +109,28 @@ def new_predefined_driven_control(scheme=PRIMITIVE, **kwargs):
...
 
     Parameters
     ----------
+    azimuthal_angle : float
+        The azimuthal position of the pulse.
     rabi_rotation : float
         The total polar angle to be performed by the pulse.
-        Defined in polar coordinates.
-    maximum_rabi_rate : float
+        Defined in polar coordinates. Defaults to None.
```
charmasaur, line 124: Remove the comment about defaulting to None

qctrlopencontrols/driven_controls/predefined.py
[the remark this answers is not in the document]
tachikoma-li: NP, this is the same issue as you commented above. Now should be fixed with making rabi_rotation required.

qctrlopencontrols/driven_controls/predefined.py
tachikoma-li: done

PR #103, opened 2020-07-28
Fixing CORPSE typo

leoadec:
Just a small fix for a typo that's been around since the beginning of time

### DIFF
--- a/qctrlopencontrols/driven_controls/__init__.py
+++ b/qctrlopencontrols/driven_controls/__init__.py
@@ -43,13 +43,13 @@ SK1 = "SK1"
 # First-order Walsh sequence control
 WAMF1 = "WAMF1"
 
-# Dynamically corrected control - Compensating for Off-Resonance with a Pulse Sequence (COPRSE)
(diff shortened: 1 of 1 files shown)


tachikoma-li:
Hopefully, we can get rid of these strings later.

leoadec:
@charmasaur @cfarrend I just merged this before I had approval from anybody from the support group (which I think are the code owners). I suspect this isn't something I should be allowed to do?

charmasaur:
Indeed, I wouldn't expect that to be allowed either. I guess this is the same thing we saw for the schemas in api2. Hopefully Chris can shed some light on it.

charmasaur:
https://docs.github.com/en/github/administering-a-repository/enabling-required-reviews-for-pull-requests suggests it's just a setting we can configure. IMO it should be turned on.

cfarrend:
Yep, it's a setting we can turn on and apply to our standards. I think it should be fine to apply across the board, I will just have a look if it conflicts with auto commits such as [this one](https://github.com/qctrl/python-open-controls/commit/b11193958a9b32c625224a93aac6220b0415474f)

PR #105, opened 2020-08-04
Add type hint for drive control module

tachikoma-li:
Add type hint

For passing `mypy` test, some refactoring is done for checking the inputs. The main purposes of these checking are:
1. set proper default values when inputs are all ``None``
2. make sure all non ``None`` inputs have the same shape
3. set the proper values for ``None`` input

Note that type hint is not added for those properties as I am planning to rearrange those checkings first. Will do in another PR.

### DIFF
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -113,10 +113,10 @@ class DrivenControl:
         if durations is None:
             durations = np.ones(input_length)
 
-        self.rabi_rates = np.array(rabi_rates)
-        self.azimuthal_angles = np.array(azimuthal_angles)
-        self.detunings = np.array(detunings)
-        self.durations = np.array(durations)
+        self.rabi_rates = np.array(rabi_rates, dtype=np.float)
+        self.azimuthal_angles = np.array(azimuthal_angles, dtype=np.float)
+        self.detunings = np.array(detunings, dtype=np.float)
+        self.durations = np.array(durations, dtype=np.float)
 
         # check if all the rabi_rates are greater than zero
         if np.any(self.rabi_rates < 0.0):
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -46,16 +46,16 @@ class DrivenControl:
     ----------
     rabi_rates : numpy.ndarray, optional
         1-D array of size nx1 where n is number of segments;
-        Each entry is the rabi rate for the segment. Defaults to None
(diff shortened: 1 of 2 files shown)


qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -65,80 +67,59 @@ class DrivenControl:
...
-                durations = np.ones((valid_input_length,))
-
-        self.rabi_rates = rabi_rates
-        self.azimuthal_angles = azimuthal_angles
-        self.detunings = detunings
-        self.durations = durations
+        # check if all non-None inputs have the same length
+        input_lengths = {
+            len(v)
+            for v in [rabi_rates, azimuthal_angles, detunings, durations]
```
charmasaur, line 89: We also used to handle the case where these were multi-dimensional (we'd check the lengths of the flattened inputs). I guess that could be relevant if the user passes `[[1,2,3]]` or something... It's a little inefficient, but the simplest way I can think of to handle that case is:
 - change the `len(v)` to `len(np.array(v).size)`
 - call `.flatten()` when creating the arrays on 116--119

tachikoma-li: SG.

> We also used to handle the case where these were multi-dimensional (we'd check the lengths of the flattened inputs). I guess that could be relevant if the user passes `[[1,2,3]]` or something... It's a little inefficient, but the simplest way I can think of to handle that case is:
> 
>     * change the `len(v)` to `len(np.array(v).size)`
IIUC, `np.array(v).size` would be fine for handling size?
>     * call `.flatten()` when creating the arrays on 116--119
SG

charmasaur: > SG.
> 
> > We also used to handle the case where these were multi-dimensional (we'd check the lengths of the flattened inputs). I guess that could be relevant if the user passes `[[1,2,3]]` or something... It's a little inefficient, but the simplest way I can think of to handle that case is:
> > ```
> > * change the `len(v)` to `len(np.array(v).size)`
> > ```
> 
> IIUC, `np.array(v).size` would be fine for handling size?

D'oh.

tachikoma-li: Done. The only issue I have now is that the docstring says all inputs are 1-D. Any reason why we want to support the multi-dimensional case here?

charmasaur: It's probably not really necessary, I think it's generally just good if wherever possible we can avoid changing functionality (in a backwards incompatible way). In this case we can manage that without too much trouble, so I think it's worth it.

PR #107, opened 2020-08-07
Add constants file

tachikoma-li:
Put all constants in a single file and change import accordingly.

### DIFF
--- a/qctrlopencontrols/constants.py
+++ b/qctrlopencontrols/constants.py
@@ -1,0 +1,109 @@
+# Copyright 2020 Q-CTRL Pty Ltd & Q-CTRL Inc
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


qctrlopencontrols/constants.py
```diff
@@ -0,0 +1,109 @@
...
+WALSH_SINGLE_AXIS = "Walsh single-axis"
+
+# Quadratic dynamical decoupling sequence
+QUADRATIC = "quadratic"
+
+# X-Concatenated dynamical decoupling sequence
+X_CONCATENATED = "X concatenated"
+
+# XY-Concatenated dynamical decoupling sequence
+XY_CONCATENATED = "XY concatenated"
```
leoadec, line 109: Were this constants taken from some other file? Is the plan to remove them from the other files as well?

tachikoma-li: These constants are copied from `__init__.py` in driven control and dynamical decoupling modules. The plan is to remove these two files after this PR.

PR #110, opened 2020-08-25
Add function to generate modulated Gaussian control

tachikoma-li:
This function should be the same as the one in core with some tweaks. We expect users to use it to generate amplitude noise characterization pulses.
Tests are migrated and modified a bit.

### DIFF
--- a/qctrlopencontrols/driven_controls/predefined.py
+++ b/qctrlopencontrols/driven_controls/predefined.py
@@ -882,16 +882,11 @@ def new_modulated_gaussian_control(
         },
     )
 
-    segment_start_times = np.arange(0, duration, minimum_segment_duration)
-    segment_num = len(segment_start_times)
-    segment_midpoints = minimum_segment_duration / 2 + segment_start_times
-
-    # prepare the modulation signals. We use sinusoids that are zero at the center of the pulse,
-    # which ensures the pulses are antisymmetric about the center of the pulse and thus effect a net
-    # zero rotation.
-    modulation_signals = np.sin(
-        2.0 * np.pi * modulation_frequency * (segment_midpoints - duration / 2)
-    )
+    # work out exact segment duration
+    segment_num = int(np.ceil(duration / minimum_segment_duration))
+    segment_duration = duration / segment_num
+    segment_start_times = np.arange(segment_num) * segment_duration
+    segment_midpoints = segment_start_times + segment_duration / 2
 
     # prepare a base gaussian shaped pulse
     gaussian_mean = _pulse_mean * duration
@@ -901,6 +896,12 @@ def new_modulated_gaussian_control(
     )
 
     if modulation_frequency != 0:
+        # prepare the modulation signals. We use sinusoids that are zero at the center of the pulse,
+        # which ensures the pulses are antisymmetric about the center of the pulse and thus effect
+        # a net zero rotation.
+        modulation_signals = np.sin(
+            2.0 * np.pi * modulation_frequency * (segment_midpoints - duration / 2)
+        )
         # modulate the base gaussian
         modulated_gaussian_segments = base_gaussian_segments * modulation_signals
 
@@ -937,5 +938,5 @@ def new_modulated_gaussian_control(
     return DrivenControl(
         rabi_rates=np.abs(modulated_gaussian_segments),
         azimuthal_angles=azimuthal_angles,
-        durations=np.array([minimum_segment_duration] * segment_num),
+        durations=np.array([segment_duration] * segment_num),
     )
--- a/qctrlopencontrols/driven_controls/predefined.py
+++ b/qctrlopencontrols/driven_controls/predefined.py
@@ -883,15 +883,15 @@ def new_modulated_gaussian_control(
     )
 
     # work out exact segment duration
-    segment_num = int(np.ceil(duration / minimum_segment_duration))
-    segment_duration = duration / segment_num
-    segment_start_times = np.arange(segment_num) * segment_duration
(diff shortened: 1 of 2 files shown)


leoadec:
I think I might be missing some context here about why we're transferring this to Open Controls?

tachikoma-li:
> I think I might be missing some context here about why we're transferring this to Open Controls?

User guide is using old API to generate noise characterization pulses, which we want to get rid of. 
To fully migrate to the new API, we want the pulse generation can be don with Open Control (like in other guides), which would save us from writing a new API for it.

qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -812,3 +813,129 @@ def _new_walsh_amplitude_modulated_filter_1_control(
...
+    )
+
+    segment_start_times = np.arange(0, duration, minimum_segment_duration)
+    segment_num = len(segment_start_times)
+    segment_midpoints = minimum_segment_duration / 2 + segment_start_times
+
+    # prepare the modulation signals. We use sinusoids that are zero at the center of the pulse,
+    # which ensures the pulses are antisymmetric about the center of the pulse and thus effect a net
+    # zero rotation.
+    modulation_signals = np.sin(
```
charmasaur, line 892: nit: how about we move this inside the `modulation_frequency != 0` block?

qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -812,3 +813,129 @@ def _new_walsh_amplitude_modulated_filter_1_control(
...
+        maximum_rabi_rate >= min_required_upper_bound,
+        "Maximum Rabi rate must be large enough to permit a 2Pi rotation.",
+        {"maximum_rabi_rate": maximum_rabi_rate},
+        extras={
+            "minimum required value for upper_bound "
+            "(sqrt(2pi)/(0.1*maximum_duration))": min_required_upper_bound
+        },
+    )
+
+    segment_start_times = np.arange(0, duration, minimum_segment_duration)
```
charmasaur, line 885: Given that it's a specified `duration` but only a *minimum* segment duration, I think instead we should use slightly longer segment lengths so that we get a pulse of duration `duration` exactly:
```
segment_count = int(np.ceil(duration / minimum_segment_duration))
segment_start_times = np.arange(segment_count) * duration / segment_count # or similar
...
```

qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -812,3 +813,129 @@ def _new_walsh_amplitude_modulated_filter_1_control(
...
+    else:
+        # for the zero-frequency pulse, we need to produce the largest possible full rotation (i.e.
+        # multiple of 2pi) while respecting the maximum Rabi rate. Note that if the maximum Rabi
+        # rate does not permit even a single rotation (which could happen to a small degree due to
+        # discretization issues) then we allow values to exceed the maximum Rabi rate.
+        normalized_gaussian_segments = base_gaussian_segments / np.max(
+            base_gaussian_segments
+        )
+        maximum_rotation_angle = (
+            minimum_segment_duration
```
charmasaur, line 922: If you make the suggested change above, this would become `segment_duration`

charmasaur (changes requested):

qctrlopencontrols/driven_controls/predefined.py
tachikoma-li: SG, as it's only used in that case.

qctrlopencontrols/driven_controls/predefined.py
tachikoma-li: That's right, sorry I missed that point. Fixed.

leoadec (commented):
The code looks good, I just have two suggestions.

I don't think I have the authority to oversee the transfer of code from a private repo to an open source project, so I'll let @charmasaur give the final thumbs up here

qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -812,3 +813,130 @@ def _new_walsh_amplitude_modulated_filter_1_control(
...
+        "Maximum Rabi rate must be large enough to permit a 2Pi rotation.",
+        {"maximum_rabi_rate": maximum_rabi_rate},
+        extras={
+            "minimum required value for upper_bound "
+            "(sqrt(2pi)/(0.1*maximum_duration))": min_required_upper_bound
+        },
+    )
+
+    # work out exact segment duration
+    segment_num = int(np.ceil(duration / minimum_segment_duration))
```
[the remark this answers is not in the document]
tachikoma-li: Yeah, somehow I always pick `segment_num`

qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -812,3 +813,130 @@ def _new_walsh_amplitude_modulated_filter_1_control(
...
+    segment_duration = duration / segment_num
+    segment_start_times = np.arange(segment_num) * segment_duration
+    segment_midpoints = segment_start_times + segment_duration / 2
+
+    # prepare a base gaussian shaped pulse
+    gaussian_mean = _pulse_mean * duration
+    gaussian_width = _pulse_width * duration
+    base_gaussian_segments = (1.0 / gaussian_width / np.sqrt(2 * np.pi)) * np.exp(
+        -0.5 * ((segment_midpoints - gaussian_mean) / gaussian_width) ** 2
+    )
```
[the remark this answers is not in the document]
tachikoma-li: SG, removed.

charmasaur (approved):

PR #115, opened 2020-09-04
Fix scaling of amplitudes when exporting controls

charmasaur:
The amplitudes and Rabi rates are supposed to be given in proportion to
the maximum Rabi rate:
https://docs.q-ctrl.com/wiki/output-data-formats#q-ctrl-hardware

Also just use the azimuthal angles directly, rather than taking an arctan of
the amplitudes.

https://q-ctrl.atlassian.net/browse/QENG-1118


tachikoma-li (approved):
Will this function be deprecated eventually?

charmasaur:
I hope so. Figured for now it's easiest just to fix it.

PR #121, opened 2020-09-22
Make predefined DD functions public

tachikoma-li:
This is part of https://q-ctrl.atlassian.net/browse/PRODUCT-2318, so we can easily check changes in doc.

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -196,25 +196,25 @@ def new_predefined_dds(scheme=SPIN_ECHO, **kwargs):
     """
 
     if scheme == RAMSEY:
-        sequence = _new_ramsey_sequence(**kwargs)
+        sequence = new_ramsey_sequence(**kwargs)
     elif scheme == SPIN_ECHO:
-        sequence = _new_spin_echo_sequence(**kwargs)
+        sequence = new_spin_echo_sequence(**kwargs)
     elif scheme == CARR_PURCELL:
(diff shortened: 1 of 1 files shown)


qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -196,25 +196,25 @@ def new_predefined_dds(scheme=SPIN_ECHO, **kwargs):
...
+        sequence = new_ramsey_sequence(**kwargs)
     elif scheme == SPIN_ECHO:
-        sequence = _new_spin_echo_sequence(**kwargs)
+        sequence = new_spin_echo_sequence(**kwargs)
     elif scheme == CARR_PURCELL:
-        sequence = _new_carr_purcell_sequence(**kwargs)
+        sequence = new_carr_purcell_sequence(**kwargs)
     elif scheme == CARR_PURCELL_MEIBOOM_GILL:
-        sequence = _new_carr_purcell_meiboom_gill_sequence(**kwargs)
+        sequence = new_carr_purcell_meiboom_gill_sequence(**kwargs)
```
charmasaur, line 205: I don't really like the fact that these names are all spelled out in full and for the driven controls they're abbreviations, but since they're consistent with the scheme names I guess it's fine to keep them.

charmasaur (approved):

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
tachikoma-li: That's a good point. I'll change them in following PRs.

PR #122, opened 2020-09-22
 Use abbreviation in predefined DD functions

tachikoma-li:
- some name tweak
- fix typo

I guess spin echo is fine? (although in the doc we do call it `se`)

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -368,7 +368,7 @@ def new_spin_echo_sequence(duration=None, pre_post_rotation=False, **kwargs):
     )
 
 
-def new_carr_purcell_sequence(
+def new_cp_sequence(
     duration=None, number_of_offsets=None, pre_post_rotation=False, **kwargs
 ):
     """
@@ -434,7 +434,7 @@ def new_carr_purcell_sequence(
     )
 
 
-def new_carr_purcell_meiboom_gill_sequence(
+def new_cpmg_sequence(
     duration=None, number_of_offsets=None, pre_post_rotation=False, **kwargs
 ):
     """
@@ -501,7 +501,7 @@ def new_carr_purcell_meiboom_gill_sequence(
     )
 
 
-def new_uhrig_single_axis_sequence(
+def new_single_axis_uhrig_sequence(
     duration=None, number_of_offsets=None, pre_post_rotation=False, **kwargs
 ):
     """
@@ -568,7 +568,7 @@ def new_uhrig_single_axis_sequence(
     )
 
 
-def new_periodic_single_axis_sequence(
+def new_single_axis_periodic_sequence(
     duration=None, number_of_offsets=None, pre_post_rotation=False, **kwargs
 ):
     """
@@ -635,7 +635,7 @@ def new_periodic_single_axis_sequence(
     )
 
 
-def new_walsh_single_axis_sequence(
+def new_single_axis_walsh_sequence(
     duration=None, paley_order=None, pre_post_rotation=False, **kwargs
 ):
     """
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -200,7 +200,7 @@ def new_predefined_dds(scheme=SPIN_ECHO, **kwargs):
     elif scheme == SPIN_ECHO:
         sequence = new_spin_echo_sequence(**kwargs)
     elif scheme == CARR_PURCELL:
-        sequence = new_cp_sequence(**kwargs)
+        sequence = new_carr_purcell_sequence(**kwargs)
     elif scheme == CARR_PURCELL_MEIBOOM_GILL:
         sequence = new_cpmg_sequence(**kwargs)
     elif scheme == UHRIG_SINGLE_AXIS:
@@ -368,7 +368,7 @@ def new_spin_echo_sequence(duration=None, pre_post_rotation=False, **kwargs):
     )
 
 
-def new_cp_sequence(
+def new_carr_purcell_sequence(
     duration=None, number_of_offsets=None, pre_post_rotation=False, **kwargs
 ):
     """
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -187,7 +187,7 @@ def new_predefined_dds(scheme=SPIN_ECHO, **kwargs):
     Returns
     ------
     qctrlopencontrols.dynamic_decoupling_sequences.DynamicDecouplingSequence
-        A dynamical decuopling sequence corresponding to `scheme`.
+        A dynamical decoupling sequence corresponding to `scheme`.
 
     Raises
     -----
@@ -200,15 +200,15 @@ def new_predefined_dds(scheme=SPIN_ECHO, **kwargs):
     elif scheme == SPIN_ECHO:
         sequence = new_spin_echo_sequence(**kwargs)
     elif scheme == CARR_PURCELL:
-        sequence = new_carr_purcell_sequence(**kwargs)
+        sequence = new_cp_sequence(**kwargs)
     elif scheme == CARR_PURCELL_MEIBOOM_GILL:
-        sequence = new_carr_purcell_meiboom_gill_sequence(**kwargs)
+        sequence = new_cpmg_sequence(**kwargs)
     elif scheme == UHRIG_SINGLE_AXIS:
-        sequence = new_uhrig_single_axis_sequence(**kwargs)
+        sequence = new_single_axis_uhrig_sequence(**kwargs)
     elif scheme == PERIODIC_SINGLE_AXIS:
-        sequence = new_periodic_single_axis_sequence(**kwargs)
+        sequence = new_single_axis_periodic_sequence(**kwargs)
     elif scheme == WALSH_SINGLE_AXIS:
-        sequence = new_walsh_single_axis_sequence(**kwargs)
+        sequence = new_single_axis_walsh_sequence(**kwargs)
     elif scheme == QUADRATIC:
         sequence = new_quadratic_sequence(**kwargs)
     elif scheme == X_CONCATENATED:
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -142,15 +142,9 @@ def _add_pre_post_rotations(
     if (preserves_10 and preserves_11) or (not preserves_10 and not preserves_11):
         final_azimuthal = np.pi
 
-    offsets = np.insert(
-        offsets,
-        [0, offsets.shape[0]],
-        [0, duration],
-    )
+    offsets = np.insert(offsets, [0, offsets.shape[0]], [0, duration],)
     rabi_rotations = np.insert(
-        rabi_rotations,
-        [0, rabi_rotations.shape[0]],
-        [rabi_value, rabi_value],
+        rabi_rotations, [0, rabi_rotations.shape[0]], [rabi_value, rabi_value],
     )
     azimuthal_angles = np.insert(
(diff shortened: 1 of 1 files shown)


qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -200,15 +200,15 @@ def new_predefined_dds(scheme=SPIN_ECHO, **kwargs):
     elif scheme == SPIN_ECHO:
         sequence = new_spin_echo_sequence(**kwargs)
     elif scheme == CARR_PURCELL:
-        sequence = new_carr_purcell_sequence(**kwargs)
+        sequence = new_cp_sequence(**kwargs)
```
perjlieb, line 203: Not sure if I follow the correct convention, but I thought two words should be spelled out, i.e., `carr_purcell` instead of `cp`.

tachikoma-li: Good point. I guess if using `cp` here, I should also change `spin echo` to `se`?
I am just so used to `CP` and `CPMG` sequences and change it without thinking too much.

charmasaur: Well, the standard specifies that three or more should be abbreviated (https://github.com/qctrl/.github/blob/master/CONTRIBUTING.md#variable-naming), but I don't think it says anything about not abbreviating fewer.

In this case I think it's just a question of how meaningful the CP abbreviation is. My understanding is that it's a fairly standard abbreviation, so I think abbreviating here is reasonable.

tachikoma-li: SGTM

charmasaur: I'd leave "spin echo" as-is too, since I don't think SE is a standard abbreviation (although feel free to correct me if I'm wrong; my evidence is just that a Google of "se dynamical decoupling" doesn't return any obvious spin echo results)

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -501,7 +501,7 @@ def new_carr_purcell_meiboom_gill_sequence(
     )
 
 
-def new_uhrig_single_axis_sequence(
+def new_single_axis_uhrig_sequence(
```
charmasaur, line 504: Google tells me that UDD is a thing -- is that the same as this, or is the "single axis" special somehow?

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
perjlieb: @charmasaur thanks for the clarification on abbreviations. That said, when I search for `CPMG` + `NMR` I get what I am looking for, but `CP` + `NMR` doesn't lead `carr purcell` but `cross polarization`.

charmasaur: OK, fair enough. If there's another CP sequence that might cause confusion with this then let's leave it expanded.

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
tachikoma-li: That's a good question. I wasn't sure about what single-axis here means in the first place. Looking at the internal document and [wiki](https://docs.q-ctrl.com/wiki/control-library), it is not documented anyway. @perjlieb  any idea about why we had this `single-axis` stated explicitly before?

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
tachikoma-li: Thanks for the comment, I'll revert the changes of `carr purcell`.

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
tachikoma-li: @michaelhush might have a better idea here.

michaelhush:
@tachikoma-li I don't understand the motivation behind these changes clearly. 

The scheme names should match the names of the methods, as they did previously.

"Carr-Purcell-Meiboom-Gill" => CARR_PURCELL_MEIBOOM_GILL => new_carr_purcell_meiboom_gill

It seems like you're making a breaking change to the API just to change some terminology. But the terminology is not necessarily wrong?

If you want to just eliminate the terms single axis and use some abbreviations. That's fine. But just do it completely consistently with the constants as well. 

So make sure you also change this file:
https://github.com/qctrl/python-open-controls/blob/master/qctrlopencontrols/constants.py

Also these changes will break the Jupyter notebook examples, wherever they ended up landing.

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
perjlieb: The single-axis refers to one of the x, y and z axes, which can be generalized to a fixed control axis (fixed phase in time). The sequence reduces the noise along this single-axis. Probably @michaelhush and @HarrisonBall can tell you more about this.

michaelhush:
@tachikoma-li if you're just migrating the documentation. I'd say just leave all the names the same, since they are currently consistent with the string names. And then just add a doc string.

tachikoma-li:
> @tachikoma-li I don't understand the motivation behind these changes clearly.
> 
> The scheme names should match the names of the methods, as they did previously.
> 
> "Carr-Purcell-Meiboom-Gill" => CARR_PURCELL_MEIBOOM_GILL => new_carr_purcell_meiboom_gill
> 
> It seems like you're making a breaking change to the API just to change some terminology. But the terminology is not necessarily wrong?
> 
> If you want to just eliminate the terms single axis and use some abbreviations. That's fine. But just do it completely consistently with the constants as well.
> 
> So make sure you also change this file:
> https://github.com/qctrl/python-open-controls/blob/master/qctrlopencontrols/constants.py
> 
> Also these changes will break the Jupyter notebook examples, wherever they ended up landing.

This PR and following PRs to add the documentation would not break any existing APIs as they only interact with `new_predefined_dds`.
To add the documentation to dynamical decoupling functions, we need to make these private functions public and I think it's a good time to make the function signatures clear and consistent.
I am not sure we need to be consistent with the string constant as `new_predefined_dds` really is just a wrapper function. And, using a string flag is not quite ideal in this case, the long term goal (in my mind) would be removing all these strings and provide an `Enum`  for `new_predefined_dds` to select different control schemes. @charmasaur in case I missed anything here.

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
tachikoma-li: Thanks for the explanation. Would UDD by default mean a single-axis sequence? If so, can we just remove the term single axis in this case (and for others)? As spin echo, CP, and CPMG are also single-axis sequences in this sense, and we don't use single-axis for them, IIUC.

tachikoma-li: https://royalsocietypublishing.org/doi/pdf/10.1098/rsta.2011.0355
From this paper, the impression I have now is a multi-axis suquence means a combination of two or more types of  rotations along differnt axis. As UDD means Y rotations applied at specific time offsets, I think it should be safe to leave the `single-axis` out (and also for `periodic` and `walsh`).

charmasaur:
I thought about this more and my take is as follows.

If we want to consider the `new_predefined_dds` method as on a path towards deprecation (in favour of `new_<foo>`) then there's no harm in using different name here, since as Li says it's not actually a breaking change, and it makes the names more consistent with the ones in driven controls, so seems a worthwhile change.

If we want to keep the `new_predefined_dds` indefinitely then I think for me the consistency of the strings with the new function names beats consistency with driven controls, so I'd vote to keep the names the same.

So maybe we can work out what we want the future of `new_predefined_dds` to be. Personally I think it should be removed eventually, since I think separate functions is strictly better, but I could be convinced the other way.

Thoughts @michaelhush @tachikoma-li ?

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
perjlieb: I think it is safe to leave `single-axis` out as well. I consulted the `numerical-package` to find out where `single-axis`  is used and it basically is a flag whether pulses have contributions to one ore more major axes (x,y, and z).

tachikoma-li: SG, I'll remove `single-axis` from function names.

michaelhush:
@charmasaur if you want to depreciate `new_predefined_dds`, which is fine, then just do it immediately.

It doesn't have to be in this PR. But I don't think it's a good idea to make something inconsistent and leave it in an odd state.

I'd also remove the `new_predefined_pulses` (or whatever they are called as well).

@tachikoma-li this point is moot now, but "I am not sure we need to be consistent with the string constant as new_predefined_dds really is just a wrapper function." Doesn't fly with me. Whatever logic you use to express why you think the new name to the methods makes more sense, must also apply to the string constants. So they should all be consistent. I.e. you're arguing that single axis is not consistent with the method names. If that's true this is also the case with the string names and should be fixed too.

So I'd say there are only two paths forward:
1. Leave the names all the same and just make the methods public.
2. Change all the names and delete all the predefined method. 

I don't mind which approach you take. For the record the predefined methods came from trying to match the needs of BLACK OPAL in some sense, where the names of the pulses where needed for displaying in the browser. With BLACK OPAL going, and also open controls not being used for that, this is not required anymore.

Making these breaking changes might create a problem with anything that depends on open controls. (i.e. the numerical package I think?) But as long as that's all version pinned that's fine.

tachikoma-li:
> I thought about this more and my take is as follows.
> 
> If we want to consider the `new_predefined_dds` method as on a path towards deprecation (in favour of `new_<foo>`) then there's no harm in using different name here, since as Li says it's not actually a breaking change, and it makes the names more consistent with the ones in driven controls, so seems a worthwhile change.
> 
> If we want to keep the `new_predefined_dds` indefinitely then I think for me the consistency of the strings with the new function names beats consistency with driven controls, so I'd vote to keep the names the same.
> 
> So maybe we can work out what we want the future of `new_predefined_dds` to be. Personally I think it should be removed eventually, since I think separate functions is strictly better, but I could be convinced the other way.
> 
> Thoughts @michaelhush @tachikoma-li ?

Thanks for the thought. To me, it is clear we should remove `new_predefined_dds` eventually for the following reasons: @charmasaur 
- it seems convenient to have a wrapper function. However, string selection is not supported by autocomplete, users still need to manually type the scheme, which is error-prone.
- as the actual doc lives in the individual functions now, for users who want to use the API, it's in fact easier for them to just call the function while reading the doc. (otherwise, they would not enjoy `shitf+Tab` in Jupyter to see the doc without leaving it)
 - The annoying keyword parameters. To make the wrapper function less cumbersome, `**kwargs` parameter is used. This has two effects:
     - it's hard to doc what these kwargs are the current doc says ` Additional keyword argument to create the sequence.`, which is not really helpful.
     - To use the `new_predefined_dds`,  users need to go through other functions signature to understand the current input parameters, which makes me feel that this wrapper function is redudent(and because of this issue, changing string selection to Enum would not fix this issue).

michaelhush:
I'm fine with losing the `predefined` methods. They historically came from stuff to do with BLACK OPAL. They don't serve any important purpose in Open Controls anymore. 

But I'd suggest you do that now, if you want to.

tachikoma-li:
> I'm fine with losing the `predefined` methods. They historically came from stuff to do with BLACK OPAL. They don't serve any important purpose in Open Controls anymore.
> 
> But I'd suggest you do that now, if you want to.

Thanks, I also think we should do it soon before too many notebooks are using `new_predefined_dds` and `new_predefined_pulses`. I am happy to check existing notebooks after making the changes.

michaelhush:
@tachikoma-li @charmasaur also in situations like this, I think it's reasonable for when you make changes to the API you have to fix them in the application notes as well. I'm not sure if any of them use DDSs at the moment. But change them if they do.

charmasaur (approved):

perjlieb (approved):

PR #126, opened 2020-10-15
Refactor `new_ramsey_sequence`

tachikoma-li:
We will apply the same pattern for other DDS functions. That is,
1. remove all `ArgumentsValueError` by using `check_arguments` function
2. remove all `kwargs`
3. fix default values

`_check_duration` will be removed after we finish refactoring all DDS functions.

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -24,6 +24,7 @@ from typing import (
 import numpy as np
 
 from ..exceptions import ArgumentsValueError
+from ..utils import check_arguments
 from .dynamic_decoupling_sequence import DynamicDecouplingSequence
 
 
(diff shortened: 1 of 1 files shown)


qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -176,30 +172,25 @@ def _check_duration(duration: Optional[float] = None) -> float:
     return duration
 
 
-def new_ramsey_sequence(duration=None, pre_post_rotation=False, **kwargs):
+def new_ramsey_sequence(duration=1, pre_post_rotation=False, name=None):
```
tachikoma-li, line 175: Shall we just set name=`ramsey` at the end while creating the `DynamicDecouplingSequence` object, and remove this input?

charmasaur: Let's keep it, that way behaviour stays the same

charmasaur (approved):

PR #129, opened 2020-10-16
Refactor `new_cpmg_sequence`

tachikoma-li:
see the context in #126

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -369,32 +369,27 @@ def new_carr_purcell_sequence(
 
 
 def new_cpmg_sequence(
-    duration=None, number_of_offsets=None, pre_post_rotation=False, **kwargs
+    duration=1.0, number_of_offsets=1, pre_post_rotation=False, name=None
 ):
     r"""
     Creates the Carr-Purcell-Meiboom-Gill sequence.
 
     Parameters
     ---------
-    duration : float
-        Total duration of the sequence :math:`\tau`. Defaults to None.
+    duration : float, optional
+        Total duration of the sequence :math:`\tau` (in seconds). Defaults to 1.
     number_of_offsets : int, optional
-        Number of offsets :math:`n`. Defaults to None.
+        Number of offsets :math:`n`. Defaults to 1.
     pre_post_rotation : bool, optional
-        If True, a :math:`X_{\pi/2}` rotation is added at the
-        start and end of the sequence.
-    kwargs : dict
-        Additional keywords required by DynamicDecouplingSequence.
+        If ``True``, a :math:`X_{\pi/2}` rotation is added at the
+        start and end of the sequence. Defaults to ``False``.
+    name : string, optional
+        Name of the sequence. Defaults to ``None``.
 
     Returns
     -------
     DynamicDecouplingSequence
         The Carr-Purcell-Meiboom-Gill sequence.
-
-    Raises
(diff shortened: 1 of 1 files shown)


qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -369,33 +369,28 @@ def new_carr_purcell_sequence(
 
 
 def new_cpmg_sequence(
-    duration=None, number_of_offsets=None, pre_post_rotation=False, **kwargs
+    duration=1.0, number_of_offsets=1, pre_post_rotation=False, name=None
```
perjlieb, line 372: Do you know if there are plans to change the parameter names in open-controls to our standard?

tachikoma-li: Good question. I don't think there is one, but I think that is reasonable to do.
I would probably fix it after the refactoring of the DDS module is done.

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -416,14 +411,15 @@ def new_cpmg_sequence(
     .. [#] `S. Meiboom and D. Gill, Review of Scientific Instruments 29:8, 688 (1958).
         <https://link.aps.org/doi/10.1063/1.1716296>`_
     """
-    duration = _check_duration(duration)
-    number_of_offsets = number_of_offsets or 1
-    number_of_offsets = int(number_of_offsets)
```
leoadec, line 421: This line `number_of_offsets = int(number_of_offsets)` used to prevent a user from mistakenly passing something like `number_of_offsets = 3.000001` which I think would cause the pulses to exceed the duration of the sequence.

I'm not entirely sure if this line is necessary, as the docstring says that the user should pass an integer for this parameter. But in any case, just wanted to know your thoughts about this

leoadec (commented):
Very nice, just have one question

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
tachikoma-li: That's a good question. In that case, would that mean `number_of_offsets` is calculated from other functions, which normally is a `float` number in python? 

The reason I removed is the doc says it should be `int`, as you mentioned. But it's python, we can't really stop people from passing a float number. Adding this line back seems to be safe. Maybe we can do that for now and ask Harry on Monday see whether there is a better way?

leoadec: Sounds good

tachikoma-li: I made the change in the private function as it seems we always calculate offset from there. In that way, we can hide this type issue from the main function.

leoadec: I'm okay with hiding it in the private function, but I think at the moment a user mistakenly passing `number_of_offsets = 3.000001` will still get offsets that go beyond the duration of the sequence, because `np.arange(3.000001)` is the same as `np.arange(4)`. Maybe we can write `spacing * np.arange(int(number_of_offsets))` in the private function?

tachikoma-li: ah, sorry, clearly I missed that part. Will fix. Thanks for pointing it out!

leoadec (approved):
Nice, I think this covers all cases

PR #130, opened 2020-10-18
Refactor `new_cpmg_sequence`

tachikoma-li:
Same as other refactor PRs.
The review is a bit boring, but I guess it would easier to track the changes.
Let me know if you think there is a better way to do it.

A discussion about `number_of_offsets = int(number_of_offsets)`
https://github.com/qctrl/python-open-controls/pull/129#discussion_r505951829
it's a minor issue, but would be good to have your thought here @charmasaur

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -451,22 +451,22 @@ def new_cpmg_sequence(
 
 
 def new_uhrig_sequence(
-    duration=None, number_of_offsets=None, pre_post_rotation=False, **kwargs
+    duration=1.0, number_of_offsets=1, pre_post_rotation=False, name=None
 ):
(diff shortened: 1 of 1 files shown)


qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -1216,6 +1217,8 @@ def _uhrig_single_axis_offsets(
     numpy.ndarray
         The offset values
     """
+    # in case a float number is passed
+    number_of_offsets = int(number_of_offsets)
```
charmasaur, line 1221: I'm fine with this since it's consistent with the other functions, but in general I prefer to do this sort of input-sanitisation as close as possible to where the user provides it. That way you limit the amount of code that can be exposed to the raw (possibly-ridiculous) user input.

charmasaur (approved):

PR #136, opened 2020-10-19
Fix default values in `new_xy_concatenated_sequence`

tachikoma-li:
The implementation needs lots of rework, but that would be another PR(s) in future.

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
@@ -921,7 +921,7 @@ def new_x_concatenated_sequence(
         duration > 0, "Sequence duration must be above zero.", {"duration": duration}
     )
     check_arguments(
-        concatenation_order > 0,
+        concatenation_order >= 1,
         "Concatenation oder must be above zero:",
         {"concatenation_order": concatenation_order},
     )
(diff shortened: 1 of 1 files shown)


qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -1025,16 +1018,17 @@ def new_xy_concatenated_sequence(
...
+    check_arguments(
+        duration > 0, "Sequence duration must be above zero.", {"duration": duration}
+    )
+    check_arguments(
+        concatenation_order > 0,
+        "Concatenation oder must be above zero:",
+        {"concatenation_order": concatenation_order},
+    )
+
     concatenation_order = int(concatenation_order)
```
charmasaur, line 1031: If `concatentation_order` is between 0 and 1 (exclusive) then the `check_argument` will pass but `concatenation_order` will still end up as zero. I'd change the check to `>= 1`.

charmasaur (changes requested):

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
tachikoma-li: fixed

charmasaur (approved):

PR #138, opened 2020-10-20
Refactor `DynamicDecouplingSequence`

tachikoma-li:
This PR mainly focuses on the value-checking of `DynamicDecouplingSequence`.

`duration` and `offsets` are now not optional as I can't find a good reason/way to do it. Note that:
 1.  following inputs, like `rabi_rotations`, depend on `offsets` if they are not provided.
 2.  use 1 and [0.5] as default values for `duration` and `offsets` just feel wrong.
 3. I think they were there because of the old design of Black Opal. Functions in this repo are obsessed to have default values all the time.

If you agree with this change, I would propose to also make duration as non-optional input for functions in the predefined module.

`number_of_offsets` is also removed as it was used for value checking.

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
@@ -34,6 +34,7 @@ from ..utils import (
     Coordinate,
     FileFormat,
     FileType,
+    check_arguments,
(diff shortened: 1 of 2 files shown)


charmasaur (approved):
Ahhh I'd always wondered about the obsession with default values in this package, but yes, compatibility with BLACK OPAL explains it. In any case, I'm 100% on board with this (and with removing any other meaningless default values).

PR #139, opened 2020-10-20
Remove unnecessary default values in DDS functions

tachikoma-li:
Part of me still wants to remove the `name` parameter as that might be there for BLACK OPAL.
That would require removing them in all driven control functions and DDS function.


charmasaur (approved):
I don't think `name` is doing any harm (unlike default values for `duration` and whatnot, which I think *were* doing harm), so I'd suggest just leaving it.

PR #140, opened 2020-10-23
Removing setup-poetry.sh

leoadec:
Currently the `poetry-setup.py` script only detects an installation of poetry if it is in version 0.12.x. However, poetry has been in version 1.x for a while now. This pull request updates the script accordingly.


charmasaur (approved):

cfarrend (commented):
Is anyone making use of this script? @qctrl/devops-engineering aren't using it so it can be deleted if no one else is

leoadec:
@cfarrend I think the only reason why this script still exists is because it is mentioned in the README  as part of the installation instructions (https://github.com/qctrl/python-open-controls#from-source). That said, I'd be happy to just update the README and delete the script

cfarrend:
This script may have been created when Poetry was still in its infancy. I think referring to the official installation instructions will be more useful https://python-poetry.org/docs/#installation as it contains instructions Linux, OSX and Windows

So yeah I'd say removing the script and suggesting the official install docs may be better for collaborators

leoadec:
Sounds good, I've removed the file and removed the mentions of it from the documentation.

charmasaur (approved):

cfarrend (approved):

PR #145, opened 2020-11-12
Change doc structure

tachikoma-li:
Use the template to automatically generate docs for all public APIs. In specific,
 - use `autosummary.rst`  and the template `qctrlopencontrol.rst` to generate the page for all APIs in a list with links to each API.
 - `autosummary_context` provides a way to list all APIs
 - `autosummary_filename_map` allows us to customize the URLs (by customizing the names of the docs)

Use the same template for class/function as that in the `python` repo to make sure the style is consistent across these repos.


charmasaur (commented):
Some high-level questions:
 - Is there any way we can avoid writing the manual list of functions/classes?
 - Are those autosummary templates custom, or the official autosummary ones? (I think we should avoid writing our own where possible.)

tachikoma-li:
> Some high-level questions:
> 
>     * Is there any way we can avoid writing the manual list of functions/classes?
I haven't figured out a way to do it, but I think it's possible. 
>     * Are those autosummary templates custom, or the official autosummary ones? (I think we should avoid writing our own where possible.)
This is modified from the python repo to make the final doc similar to what we have for `Qctrl`.

tachikoma-li:
The page for class seems to be working now. @charmasaur

charmasaur:
A few more things to keep in mind:
 - restructuring the docs comes with a maintenance penalty, since we have to update all references to them (and need to synchronize those updates across multiple repos); we therefore want to do it as few times as possible. That's not an argument against doing it at all, but it is an argument against making small iterative improvements.
 - ideally we want the URLs, breadcrumbs, and left hand side nav bar to be consistency structured. At the moment in this PR the URLs are structured based on the Python objects, but the breadcrumbs and nav bar have the "driven controls" / "dynamical decoupling" split. Having such a split is fine, but it shouldn't be reflected in the actual structure of the site, which should instead follow the URLs / Python objects (for an example see the "API reference" vs "Deprecated modules" split here: https://docs.q-ctrl.com/references/python/qctrl/index.html).
 - ideally we want the URLs to be consistent with how they're set up in the `qctrl` package, e.g. we'd want docs.q-ctrl.com/references/python/qctrl-open-controls/new_carr_purcell_sequence.html.

I don't really know how hard it is to make the latter two happen. They work reasonably well in the `qctrl` package, but there's a lot of custom handling there, which it would be nice to avoid if we can (although I'd guess it's not possible).

tachikoma-li:
> A few more things to keep in mind:
> 
>     * restructuring the docs comes with a maintenance penalty, since we have to update all references to them (and need to synchronize those updates across multiple repos); we therefore want to do it as few times as possible. That's not an argument against doing it at all, but it is an argument against making small iterative improvements.
> 
>     * ideally we want the URLs, breadcrumbs, and left hand side nav bar to be consistency structured. At the moment in this PR the URLs are structured based on the Python objects, but the breadcrumbs and nav bar have the "driven controls" / "dynamical decoupling" split. Having such a split is fine, but it shouldn't be reflected in the actual structure of the site, which should instead follow the URLs / Python objects (for an example see the "API reference" vs "Deprecated modules" split here: https://docs.q-ctrl.com/references/python/qctrl/index.html).
> 
>     * ideally we want the URLs to be consistent with how they're set up in the `qctrl` package, e.g. we'd want docs.q-ctrl.com/references/python/qctrl-open-controls/new_carr_purcell_sequence.html.
> 
> 
> I don't really know how hard it is to make the latter two happen. They work reasonably well in the `qctrl` package, but there's a lot of custom handling there, which it would be nice to avoid if we can (although I'd guess it's not possible).

Thanks for the tips! I confess I totally forget the URLs part. Will think about it.
I would suggest to keep this branch (can close this PR for now) and iterate up to the point we are all happy with it. At 
the moment, I can think of two things:
1. make the categorization automatic
2. the URLs issue

tachikoma-li:
Try to use a template to automatically generate the doc, using the tricks from the `python` repo.
URLs now (hopefully) are consistent as those generated from the `python` repo, IIUC.

charmasaur:
Looking better, but there's still some weirdness (at least when I build):
 - the nav bar on the left changes between different pages (from "Installation" it shows "API REFERENCE"; for any of the classes/functions it just shows "SETUP")
 - I don't think it really makes sense to separate classes and functions that are in the same module (we don't do this for the `qctrl` docs)
 - there's still extra `qctrlopencontrols` parts of the path and object name on class/function pages (e.g. `.../qctrlopencontrols/qctrlopencontrols.DynamicDecouplingSequence.html` should be `.../DynamicDecouplingSequence.html`)

At a higher level I'm also not really sure how we should be structuring things. Pure consistency with how the user writes code would lead to something like:
```
SETUP
  <whatever we want>
API REFERENCE
  new_bb1_control
  new_carr_purcell_sequence
  ...
```

I suppose we could also do something like:
```
SETUP
  <whatever we want>
API REFERENCE
  Driven controls ^
    new_bb1_control
    ...
  Dynamical decoupling sequences ^
    new_carr_purcell_sequence
    ...
```
but I *don't* think "Driven controls" and "Dynamical decoupling sequences" should be linkified in that case, because (similar to "API REFERENCE" and "DEPRECATED MODULES" in `qctrl`) it's an arbitrary split, so doesn't make sense to be reflected in the actual structure of the website.

IMO the first is simpler, although what you've got is reasonably close to the second. I guess with that in mind, keeping the structure you have is fine even if you can't unlinkify the "Driven controls" and "Dynamical decoupling sequences" parts, but we just need to try to remember not to link to those from user guides, because they're not "real" sections (given how we've exported the functions/classes from the package).

Anyway, I imagine the comments above are probably annoying. But there are two large negatives we get by doing this (changing the URLs and adding a whole lot of custom code that we need to maintain), so in order for it to be worth it I think we need to make sure that what we end up with is good enough that we won't want to change it for a while.

charmasaur (commented):
(see above)

tachikoma-li:
> Looking better, but there's still some weirdness (at least when I build):
> 
>     * ~the nav bar on the left changes between different pages (from "Installation" it shows "API REFERENCE"; for any of the classes/functions it just shows "SETUP")~ (Edit: ugh, non-clean build)
> 
>     * I don't think it really makes sense to separate classes and functions that are in the same module (we don't do this for the `qctrl` docs)
> 
>     * there's still extra `qctrlopencontrols` parts of the path and object name on class/function pages (e.g. `.../qctrlopencontrols/qctrlopencontrols.DynamicDecouplingSequence.html` should be `.../DynamicDecouplingSequence.html`)
> 
> 
> At a higher level I'm also not really sure how we should be structuring things. Pure consistency with how the user writes code would lead to something like:
> 
> ```
> SETUP
>   <whatever we want>
> API REFERENCE
>   new_bb1_control
>   new_carr_purcell_sequence
>   ...
> ```
> 
> I suppose we could also do something like:
> 
> ```
> SETUP
>   <whatever we want>
> API REFERENCE
>   Driven controls ^
>     new_bb1_control
>     ...
>   Dynamical decoupling sequences ^
>     new_carr_purcell_sequence
>     ...
> ```

> but I _don't_ think "Driven controls" and "Dynamical decoupling sequences" should be linkified in that case, because (similar to "API REFERENCE" and "DEPRECATED MODULES" in `qctrl`) it's an arbitrary split, so doesn't make sense to be reflected in the actual structure of the website.

Thanks for the feedback! One of the reasons I had `Class` and `Function` here is I was unable to combine the class/function from different files in the same module (e.g. the driven_control and predefined file in the driven control module). 
Now I find a way to get round it: the autosummary template now generates two types of docs, one to combine the APIs for a single module (with the help of use autosummary_context) and the other to generate the acutal doc.

The doc should now look as the second form you proposed above.  The URLs should behave as what we had for the BO APIs (at least works fine locally).

> IMO the first is simpler, although what you've got is reasonably close to the second. I guess with that in mind, keeping the structure you have is fine even if you can't unlinkify the "Driven controls" and "Dynamical decoupling sequences" parts, but we just need to try to remember not to link to those from user guides, because they're not "real" sections (given how we've exported the functions/classes from the package).
> 
> Anyway, I imagine the comments above are probably annoying. But there are two large negatives we get by doing this (changing the URLs and adding a whole lot of custom code that we need to maintain), so in order for it to be worth it I think we need to make sure that what we end up with is good enough that we won't want to change it for a while.

I totally agree we need to make this as good as we can such that we don't need to change it for a long time, so I appreciate your time and suggestion here.

docs/autosummary.rst
```diff
@@ -0,0 +1,33 @@
+.. Automatically generates summaries and documentation.
```
charmasaur, line 1: I confess I don't know what's going on in this file. Why are there two entries each for driven controls and dynamical decoupling sequences? Some comments could be helpful.

docs/autosummary.rst
```diff
@@ -0,0 +1,33 @@
...
+  .. autosummary::
+    :toctree: .
+    :template: drivencontrol/drivencontrol.rst
+
+    driven_controls
+
+.. automodule:: qctrlopencontrols
+  .. autosummary::
+    :toctree: .
+    :template: drivencontrol/drivencontrol.rst
```
charmasaur, line 13: If we're using this template here then it should be renamed.

docs/autosummary.rst
```diff
@@ -0,0 +1,33 @@
...
+    :template: drivencontrol/api.rst
+
+    driven_control
+    predefined
+
+.. automodule:: qctrlopencontrols.dynamic_decoupling_sequences
+  .. autosummary::
+    :toctree: .
+    :nosignatures:
+    :template: drivencontrol/api.rst
```
charmasaur, line 30: This directory needs to be renamed too

charmasaur (commented):
Getting there. Regarding delinkifying the headings, I've got no idea -- I've tried to stay as far away from Sphinx as I can, especially from customising Sphinx. If we can't do it then it's not a big deal.

The URLs still aren't quite right: .../qctrlopencontrols/DynamicDecouplingSequence.html should be .../DynamicDecouplingSequence.html (because "qctrl-open-controls" is already going to be in the URL; see https://docs.q-ctrl.com/references/python/qctrl/Operations/abs.html, for example).

I'm also not sure what to do about the breadcrumbs (https://romper-image.appspot.com/img/ag5zfnJvbXBlci1pbWFnZXIWCxIJSW1hZ2VJdGVtGICAgPTP04cKDA.png). At the moment they're consistent with the navigation bar on the left, but inconsistent with the URLs. IMO if possible we should make them consistent with the URLs (and then only the navigation bar will be "wrong"), so in the image above it would just be "Docs > new_cpmg_sequence".

docs/autosummary.rst
tachikoma-li: This is a good point. The first `autosummary` generates the qctrlopencontrol page as a big list, the rest two generate docs for DD and DDS modules. I manually added sub-module names for DD and DDS and that is why there are two entries before. It turns out we can add the `:recursive:` option to make it automatic.

docs/autosummary.rst
tachikoma-li: changed.

docs/autosummary.rst
tachikoma-li: done.

tachikoma-li:
There is now a `source` link in the doc. This is almost for free by only adding the `sphinx.ext.viewcode` extention. Let me know whether you think it's necessary.

charmasaur (commented):
OK, this is looking good now, the next question is how much of this new code do we really need -- in general we want the smallest amount of custom code possible:
 - can we get rid of the driven controls and DDS stuff in autosummary.rst?
 - do we actually need the `autosummary_context` and `autosummary_filename_map`? I guess the latter is what changes the path fragments from "qctrlopencontrols/qctrlopencontrols.foo" to "qctrlopencontrols/foo", and if that's correct then it's fine to keep it. But what about `autosummary_context`? I would have thought that could be generated automatically, in the `members` field of the templates, or something like that...
 - do we need all the templates we have?
 - and so on.

tachikoma-li:
> OK, this is looking good now, the next question is how much of this new code do we really need -- in general we want the smallest amount of custom code possible:
> 
>     * can we get rid of the driven controls and DDS stuff in autosummary.rst?
> 
>     * do we actually need the `autosummary_context` and `autosummary_filename_map`? I guess the latter is what changes the path fragments from "qctrlopencontrols/qctrlopencontrols.foo" to "qctrlopencontrols/foo", and if that's correct then it's fine to keep it. But what about `autosummary_context`? I would have thought that could be generated automatically, in the `members` field of the templates, or something like that...

AFAIK, `autosummary_context` is for mapping the URL, which I learn from `python`. 
None of the [default variables](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html) in `autosummary` works like `autosummary_context`, to my knowledge. But I could be wrong since I only dive deep into sphinx for this PR recently.
The idea here is to use the DD and DDS in autosummary.rst to generate the individual doc for all public APIs and use qctrlopencontrol.rst to list all of them. In this way, I can't see how to do it without having DD and DDS in autosummary.rst

>     * do we need all the templates we have?
module.rst is now removed. function and class are used for the same style in python repo.
qctrlopencontrol template is for group class/function in a single list.
> 
>     * and so on.

tachikoma-li:
tweak the class template to make it similar to that in the `python` repo and therefore need to add the method/attribute template.
Note that the URL for the class methods/attributes is not quite right (with an extra qctrlopencontrol), since I didn't do the map in autosummary_file_map. Would it be a problem? (although I would think there is a rare chance we need to refer to method/attribute URLs directly)

docs/conf.py
```diff
@@ -89,3 +102,32 @@
     "https://docs.q-ctrl.com/assets/js/readthedocs.js",
 ]
 html_logo = "_static/logo.svg"
+
+
+public_apis = qctrlopencontrols.__all__
+
+autosummary_context = {
```
charmasaur, line 109: Would you mind adding some brief comments explaining what each of these blocks is for, for the reader who's unfamiliar with autosummary (specifically `autosummary_context`, the `autosummary_filename_map`, and the class members handling)?

charmasaur (commented):
Very nice, looking really good now. Just a request for some comments, so that we can snapshot your current understanding of how these things work and avoid having to work it out again in the future!

Also, could you update the PR description based on the new approach?

tachikoma-li:
> Very nice, looking really good now. Just a request for some comments, so that we can snapshot your current understanding of how these things work and avoid having to work it out again in the future!
> 
> Also, could you update the PR description based on the new approach?

I added doc in `conf.py` and updated the PR description. Let me know if you want to add more comments to other files.

charmasaur (approved):

PR #148, opened 2020-11-20
Remove default values of DDS

tachikoma-li:
Remove unnecessary default values for the DDS class.
Also, remove the check for the number of offsets exceeds `UPPER_BOUND_OFFSETS` as I think this was for BLACK OPAL.

I think we can turn DDS into a dataclass and just remove the `__repr__` and `__str__`, which seem quite unnecessary here.

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
@@ -50,9 +50,9 @@ class DynamicDecouplingSequence:
         The times offsets :math:`\{t_j\}` in seconds for the center of pulses.
     rabi_rotations : np.ndarray
         The rabi rotation :math:`\omega_j` at each time offset :math:`t_j`.
-    azimuthal_angles : np.ndarray, optional
+    azimuthal_angles : np.ndarray
         The azimuthal angle :math:`\phi_j` at each time offset :math:`t_j`.
-    detuning_rotations : np.ndarray, optional
+    detuning_rotations : np.ndarray
         The detuning rotation :math:`\delta_j` at each time offset :math:`t_j`.
     name : str, optional
         Name of the sequence. Defaults to None.
--- a/qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
@@ -25,7 +25,6 @@ import numpy as np
 
 from ..constants import (
     UPPER_BOUND_DETUNING_RATE,
-    UPPER_BOUND_OFFSETS,
     UPPER_BOUND_RABI_RATE,
 )
 from ..driven_controls.driven_control import DrivenControl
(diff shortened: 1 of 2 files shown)


qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
```diff
@@ -49,18 +48,12 @@ class DynamicDecouplingSequence:
         The total time in seconds for the sequence :math:`\tau`.
     offsets : np.ndarray
         The times offsets :math:`\{t_j\}` in seconds for the center of pulses.
-    rabi_rotations : np.ndarray, optional
+    rabi_rotations : np.ndarray
         The rabi rotation :math:`\omega_j` at each time offset :math:`t_j`.
-        Defaults to ``None``.
-        If not provided, it will be set as :math:`\pi` at each time offset.
     azimuthal_angles : np.ndarray, optional
```
charmasaur, line 53:
```suggestion
    azimuthal_angles : np.ndarray
```

qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
```diff
@@ -49,18 +48,12 @@ class DynamicDecouplingSequence:
...
         The azimuthal angle :math:`\phi_j` at each time offset :math:`t_j`.
-        Defaults to ``None``.
-        If not provided, it will be set as 0 at each time offset.
     detuning_rotations : np.ndarray, optional
```
charmasaur, line 55:
```suggestion
    detuning_rotations : np.ndarray
```

qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
```diff
@@ -86,66 +79,51 @@ def __init__(
...
-            "Number of offsets is above the allowed number of maximum offsets. ",
-            {"offsets": offsets},
-            extras={
-                "number_of_offsets": len(offsets),
-                "allowed_maximum_offsets": UPPER_BOUND_OFFSETS,
-            },
-        )
-        check_arguments(
-            np.all(offsets >= 0.0) and np.all(offsets <= duration),
+            np.all((offsets >= 0) & (duration >= offsets)),
```
charmasaur, line 96:
```suggestion
            np.all(0 <= offsets <= duration),
```
Does that work?

charmasaur (changes requested):

qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
tachikoma-li: nope, python would complain `The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`,
and I was quite surprised.

charmasaur: Ah, it's equivalent to `0 <= offsets and offsets <= duration` (https://docs.python.org/2.3/ref/comparisons.html), which would explain it. Oh well.

charmasaur (approved):

PR #149, opened 2020-11-20
Update .gitignore files

tachikoma-li:
ignore the intermediates files while generating doc locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,8 @@ instance/
 
 # Sphinx documentation
 docs/_build/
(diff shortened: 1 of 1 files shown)


charmasaur (approved):

charmasaur:
It'd be nice if these could get generated into `_build`, but I don't know if there's actually a way to make that happen.

tachikoma-li:
> It'd be nice if these could get generated into `_build`, but I don't know if there's actually a way to make that happen.

That's true. I'll think about it. If possible, might change it back later.

PR #71, opened 2020-04-08
Add example notebooks

charmasaur:
https://q-ctrl.atlassian.net/browse/QENG-802
#70

PR #151, opened 2020-11-29
Update driven control notebook

tachikoma-li:
In preparation of removing `new_primitive_control`, here we use specicifc functions to create driven controls.

Note that the the values of `amplitude_x` , `amplitude_y` are different from old values. Looking at the history,  the reason is we changed how these values are calculated in #115. The notebooks were added in  #71, but not updated after merging #115.


examples/creating-a-driven-control.ipynb
```diff
@@ -40,17 +46,17 @@
...
-    "1. `primitive`\n",
-    "2. `BB1`\n",
-    "3. `SK1`\n",
-    "4. `CORPSE`\n",
-    "5. `WAMF1`\n",
-    "6. `SCROFULOUS`\n",
-    "7. `COPRSE in BB1`\n",
-    "8. `CORPSE in SK1`\n",
-    "9. `CORPSE in SCROFULOUS`\n",
+    "1. [`new_primitive_control`](https://docs.q-ctrl.com/references/python/qctrl-open-controls/qctrlopencontrols/new_primitive_control.html)\n",
```
charmasaur, line 49:
```suggestion
    "1. [Primitive](https://docs.q-ctrl.com/references/python/qctrl-open-controls/qctrlopencontrols/new_primitive_control.html)\n",
```

IMO it reads a little strange to say "... according to the following dynamically corrected gate protocols: new_primitive_control, new_bb1_control, ...", because "new_primitive_control" isn't a dynamically corrected gate protocol, it's a function. I'd suggest changing the list items like this.

tachikoma-li: Good point. Fixed now.

charmasaur (approved):

PR #153, opened 2020-12-01
Remove bound constraints on driven control

tachikoma-li:
This PR proposes to remove `bound constraints` defined in https://github.com/qctrl/python-open-controls/blob/master/qctrlopencontrols/constants.py
These values don't have particular physical meaning, but might be for the consideration of Black Opal. Since we don't have this constrain now, it might be good to remove these bounds and let user decide what to do with bounds depnding on the limitations of their apparatus.

The draft PR only removes the constraints in `DrivenControl` class. Would remove them from the whole repo if you think this is a sensible proposal.

### DIFF
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -20,13 +20,6 @@ from typing import Optional
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


charmasaur (approved):
Seems reasonable to me

PR #154, opened 2020-12-02
Remove bounds of controls

tachikoma-li:
As discussed in #153, we now remove all constant bounds in this repo.

### DIFF
--- a/tests/test_driven_controls.py
+++ b/tests/test_driven_controls.py
@@ -23,11 +23,6 @@ import numpy as np
 import pytest
 
 from qctrlopencontrols import DrivenControl
-from qctrlopencontrols.constants import (
-    UPPER_BOUND_DETUNING_RATE,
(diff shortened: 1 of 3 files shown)


tests/test_driven_controls.py
```diff
@@ -131,9 +126,6 @@ def test_driven_controls():
         _ = DrivenControl(rabi_rates=[-1])
```
charmasaur, line 126: Not for this PR (and I know it wasn't you that added this test), but note that this approach to checking multiple calls for errors doesn't work -- the test will pass if *any* of the lines throw an exception.

charmasaur (approved):

tests/test_driven_controls.py
tachikoma-li: yes, this wrong pattern has been used a lot in the tests. We need to fix them in a seperate PR later.

PR #155, opened 2020-12-03
Remove default values in `convert_dds_to_driven_control`

tachikoma-li:
- remove unnecessary default values in convert_dds_to_driven_control
- refactor the checking function a bit
-  change default value of `minimum_segment_duration` to `None`. If it's `None`, set it to 0 in the code such that we don't change the behaviour of this function

### DIFF
--- a/qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
@@ -269,7 +269,7 @@ def convert_dds_to_driven_control(
     maximum_rabi_rate: float,
     maximum_detuning_rate: float,
     minimum_segment_duration: float = 0.0,
-    name=Optional[str],
+    name: Optional[str] = None,
 ) -> DrivenControl:
     r"""
     Creates a Driven Control based on the supplied DDS and other relevant information.
--- a/qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
+++ b/qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
@@ -268,7 +268,7 @@ def convert_dds_to_driven_control(
     dynamic_decoupling_sequence: DynamicDecouplingSequence,
     maximum_rabi_rate: float,
     maximum_detuning_rate: float,
-    minimum_segment_duration: Optional[float] = None,
+    minimum_segment_duration: float = 0.0,
     name=Optional[str],
 ) -> DrivenControl:
     r"""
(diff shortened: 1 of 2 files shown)


qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
```diff
@@ -265,35 +265,32 @@ def export_to_file(
...
 def convert_dds_to_driven_control(
-    dynamic_decoupling_sequence: DynamicDecouplingSequence = None,
-    maximum_rabi_rate: float = 2 * np.pi,
-    maximum_detuning_rate: float = 2 * np.pi,
-    minimum_segment_duration: float = 0.0,
-    **kwargs,
+    dynamic_decoupling_sequence: DynamicDecouplingSequence,
+    maximum_rabi_rate: float,
+    maximum_detuning_rate: float,
+    minimum_segment_duration: Optional[float] = None,
```
charmasaur, line 271: Hm, IMO it would be simpler just to default this to zero, and even to allow the user to pass zero -- zero seems like a reasonably valid value to me.

tachikoma-li: SG, change back to zero.

qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
```diff
@@ -265,35 +265,32 @@ def export_to_file(
...
 def convert_dds_to_driven_control(
-    dynamic_decoupling_sequence: DynamicDecouplingSequence = None,
-    maximum_rabi_rate: float = 2 * np.pi,
-    maximum_detuning_rate: float = 2 * np.pi,
+    dynamic_decoupling_sequence: DynamicDecouplingSequence,
+    maximum_rabi_rate: float,
+    maximum_detuning_rate: float,
     minimum_segment_duration: float = 0.0,
-    **kwargs,
+    name=Optional[str],
```
charmasaur, line 272:
```suggestion
    name: Optional[str] = None,
```
sorry, missed this one.

tachikoma-li: Thanks!
I thought `mypy` or `pylint` would complain it.

charmasaur (approved):

qctrlopencontrols/dynamic_decoupling_sequences/dynamic_decoupling_sequence.py
charmasaur: Haha, I was surprised too, but then when I thought about it I realised that according to those there's nothing wrong: `name` is an untyped parameter for which you're assigning the value `Optional[str]` as the default.

PR #156, opened 2020-12-07
Remove default values in DrivenControl

tachikoma-li:
- remove unnecessary default values in `DrivenControl`
- modify tests depending on the default values
- add `detunings` to `new_modulated_gaussian_control` as defaults values  are now removed.

### DIFF
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -36,29 +36,22 @@ class DrivenControl:
 
     Parameters
     ----------
-    rabi_rates : np.ndarray, optional
+    rabi_rates : np.ndarray
         The Rabi rates :math:`\{\Omega_n\}` for each segment, in units of radians per second. Every
         element must be non-negative. Represented as a 1D array of length :math:`N`, where :math:`N`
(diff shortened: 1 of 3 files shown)


qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -36,30 +36,23 @@ class DrivenControl:
 
     Parameters
     ----------
-    rabi_rates : np.ndarray, optional
+    rabi_rates : np.ndarray
```
charmasaur, line 39: Hmm, unlike a lot of the other default values that you're (rightfully) getting rid of, I actually think `rabi_rates`, `azimuthal_angles`, and `detunings` defaulting to zero actually makes some sense. It would be quite common to have a control that only sets `rabi_rates`, for example, and in that case it's nice to save the user the trouble of creating zeros manually.

I agree that `durations` shouldn't have a default though.

tachikoma-li: Good point. Now only remove default values for `detunings.

We actually use `pi` as the default value for `rabi_rates` when all inputs are `None` and this should be fixed now as  `duration` is not allowed to be `None`.

Also, add the test back.

charmasaur (approved):

PR #158, opened 2020-12-09
Remove default values in `new_bb1_control`

tachikoma-li:


qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -183,8 +180,8 @@ def new_bb1_control(
         Defaults to :math:`2\pi`.
```
charmasaur, line 180: Remove

qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -164,9 +161,9 @@ def new_primitive_control(
 
 def new_bb1_control(
     rabi_rotation: float,
-    azimuthal_angle: float = 0.0,
-    maximum_rabi_rate: float = 2.0 * np.pi,
-    **kwargs
+    azimuthal_angle: float,
```
charmasaur, line 164: The docs still state that this is optional. Either update the docs, or keep this defaulting to zero (I think that would be quite reasonable, FWIW).

qctrlopencontrols/driven_controls/predefined.py
tachikoma-li: removed

qctrlopencontrols/driven_controls/predefined.py
tachikoma-li: Agreed. I also add 0 back as the default value of  `azimuthal_angle`  in `new_primitive_control`

charmasaur (approved):

PR #164, opened 2020-12-18
misc: Remove default value in `new_wamf1_control`

tachikoma-li:
Should be the last one of the `default-value` series.
Thanks @charmasaur !


charmasaur (approved):
Sorry, missed this one in the list of reviews

PR #173, opened 2021-02-11
fix: Documentation string math

anuragm:
Some minor math fixes in the comment string.

### DIFF
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -69,7 +69,7 @@ class DrivenControl:
                \frac{1}{2}\Delta(t)\sigma_z,
 
     where :math:`\Omega(t)` is the Rabi rate, :math:`\phi(t)` is the azimuthal angle (or drive
-    phase), :math:`\Delta(t)` is the detuning, :math:`\sigma_\pm = (\sigma_x \mp \sigma_y)/2`,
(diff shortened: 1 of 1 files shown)


charmasaur (approved):

qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -65,11 +65,11 @@ class DrivenControl:
     .. math::
 
         H(t) = \frac{1}{2}\left(\Omega(t) e^{i\phi(t)} \sigma_- +
-                                \Omega(t) e^{-i\phi(t)}\sigma_+\right) +
+                                \Omega^{*}(t) e^{-i\phi(t)}\sigma_+\right) +
```
charmasaur, line 68: Omega is real here, the phase phi is treated separately

anuragm: Duh, correct.

charmasaur (approved):

qctrlopencontrols/driven_controls/driven_control.py
anuragm: Force pushed to correct this, it's too minor for another commit.

charmasaur: Note that since we squash merge into master it doesn't really matter how messy the history in the branch gets.

leoadec (approved):

tachikoma-li (approved):

PR #179, opened 2021-04-13
feat: Enforce non-negativity of Rabi rotations

tiborr:
Changes proposed in this pull request:

- Enforce non-negativity of Rabi rotations during creation of a `DynamicDecouplingSequence` object

PR #181, opened 2021-04-14
misc: Add check for PR names

tachikoma-li:
To ensure we have consistent PR names as discussed in #179

### DIFF
--- a/.github/workflows/check-pr-format.yml
+++ b/.github/workflows/check-pr-format.yml
@@ -1,0 +1,22 @@
+name: Check PR format workflow
+
+on:
+  pull_request:
+    types: [opened, reopened, edited]
+    branches: [master]
+
+jobs:
+  gitlint:
+    # Filter out Draft Pull Requests
+    if: github.event.pull_request.draft == false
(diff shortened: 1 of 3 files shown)


leoadec (approved):
I'm okay with this, but I'll let @cfarrend decide on the strategy he prefers for adopting gitlint across the repos.

tachikoma-li:
> I'm okay with this, but I'll let @cfarrend decide on the strategy he prefers for adopting gitlint across the repos.

ah, true, I forgot that part. If there is already a plan about applying `gitlint` across all repos, I am happy to close the PR.

leoadec:
@tachikoma-li I'm not sure if we already have a plan, but devops have been thinking about it, so they might be the best to advise on which approach will be easier to maintain in the future.

tachikoma-li:
> @tachikoma-li I'm not sure if we already have a plan, but devops have been thinking about it, so they might be the best to advise on which approach will be easier to maintain in the future.

I guess I'll merge this one for now. We can revise it once dev has a better way to consistently manage this issue across repos.

Issue #92, opened 2020-07-09
Implement predefined Gaussian driven control


charmasaur:
Gaussian pulses are popular in quantum control, since they have a constrained bandwidth (which helps mitigate the effect of leakage).

This task is to implement a predefined Gaussian driven control. I imagine the parameters we'll need are duration, total Rabi rotation (which will determine the scaling of the Gaussian), number of segments (controlling the piecewise-constant approximation of the ideal smooth Gaussian), and width (giving the standard deviation of the Gaussian). The output should be a DrivenControl describing a piecewise-constant Gaussian envelope centered in the middle of the provided duration, with the specified width, and with amplitude chosen so that the integral of the pulse equals the specified Rabi rotation.

See https://github.com/qctrl/python-open-controls/blob/master/qctrlopencontrols/driven_controls/predefined.py for the other predefined driven controls.

charmasaur:
Here's the paper introducing DRAG, which also has a bit of a discussion about Gaussian pulses: https://arxiv.org/pdf/0901.0534.pdf

tachikoma-li:
Closed as implemented in #182.

PR #182, opened 2021-04-19
feat: Add predefined Gaussian driven control

tiborr:

- Add predefined Gaussian driven control as specified in #92


tests/test_predefined_driven_controls.py
```diff
@@ -550,6 +551,33 @@ def test_walsh_control():
...
+
+    # compute total rotation of generated pulse
+    rabi_rotation = np.dot(gaussian_control.rabi_rates, gaussian_control.durations)
+
+    # check number and duration of pulses
+    assert len(gaussian_control.rabi_rates) == _segment_count
+    assert np.allclose(gaussian_control.durations, _duration / _segment_count)
+
+    # check total rotation of pulse
+    assert np.isclose(rabi_rotation, _rabi_rotation)
```
leoadec, line 578: What if we also tested that the shape of controls matches the expected Gaussian, just like the next test checks that the `segment_values` are all close to the `expected_gaussian`?

leoadec (commented):
Thanks for your contribution, as far as I can tell your implementation matches the description of the function requested in the open issue. I just have a suggestion of an additional unit test.

tests/test_predefined_driven_controls.py
tiborr: Thank you! Unfortunately, I think one has to discretize the Gaussian first anyways to determine the proper normalization factor, so it ends up looking quite similar to the implementation itself. But would something like this work?
```suggestion
    # determine the segment mid-points
    segment_mid_points = _duration / _segment_count * (0.5 + np.arange(_segment_count))
    base_gaussian_mean = _duration * 0.5
    base_gaussian_width = _width
    base_gaussian = np.exp(
        -0.5 * ((segment_mid_points - base_gaussian_mean) / base_gaussian_width) ** 2.0
    )

    # translate pulse by B/A (from Motzoi paper) to ensure output is 0 at t=0
    y_translation = -np.exp(-0.5 * ((0 - base_gaussian_mean) / base_gaussian_width) ** 2)
    base_gaussian += y_translation
    base_gaussian_total_rotation = np.sum(base_gaussian) * _duration / _segment_count
    expected_gaussian = (
        base_gaussian / base_gaussian_total_rotation
    ) * _rabi_rotation

    # check number and duration of pulses
    assert len(gaussian_control.rabi_rates) == _segment_count
    assert np.allclose(gaussian_control.durations, _duration / _segment_count)

    # check total rotation of pulse
    assert np.isclose(rabi_rotation, _rabi_rotation)
    
    # check shape of controls matches the expected Gaussian
    assert np.allclose(gaussian_control.rabi_rates, expected_gaussian)
```

tachikoma-li (commented):
Thanks for the PR. Implementation looks good.

The test is a bit tricky since we don't want to use the same logic as the implementation in the test. I guess one possible way to do it is to work out A and B explicitly for a NOT gate with a specific `t_g` and then check the values using Eq 5 in the paper?

tiborr:
Thanks, this would work. Just a note that I think even after finding A and B, we'll still need to rescale the output of Eq. 5 before comparing with the values of the segments using code similar to the implementation itself because the segments are scaled such that their discrete integral is equal to `rabi_rotation`, not such that they actually take on values of the ideal Gaussian (i.e. the Gaussian normalised such that its integral is `rabi_rotation`).

tests/test_predefined_driven_controls.py
leoadec: That's a good point, reproducing the logic of the function in the test is probably not ideal. As @tachikoma-li mentioned below, we can probably replace A and B with the expected values in this example. I'd go further and say we probably can eliminate A from the check that is confirming that the shape is Gaussian, as the value of A is already being checked by `assert np.isclose(rabi_rotation, _rabi_rotation)`. In this case, maybe we can use something like the following as the extra check:
```
times = np.linspace(0.0005, 0.0095, 10)
gauss = lambda t: np.exp(-0.5*((t-_duration/2)/_width)**2)

expected_normalized_pulse = (gauss(times) - gauss(0))/max(gauss(times) - gauss(0))
normalized_pulse = gaussian_control.rabi_rates/max(gaussian_control.rabi_rates)

assert np.allclose(expected_normalized_pulse, normalized_pulse)
```

tiborr: Good idea, decoupling the shape check from the normalization check would be better.

```suggestion
    _segment_width = _duration / _segment_count
    midpoints = np.linspace(
        _segment_width / 2, _duration - _segment_width / 2, _segment_count
    )

    def gauss(t):
        return np.exp(-0.5 * ((t - _duration / 2) / _width) ** 2)

    expected_normalized_pulse = (gauss(midpoints) - gauss(0)) / max(
        gauss(midpoints) - gauss(0)
    )
    normalized_pulse = gaussian_control.rabi_rates / max(gaussian_control.rabi_rates)

    # check pulse is Gaussian-shaped
    assert np.allclose(expected_normalized_pulse, normalized_pulse)

    # compute total rotation of generated pulse
    rabi_rotation = np.dot(gaussian_control.rabi_rates, gaussian_control.durations)

    # check number and duration of pulses
    assert len(gaussian_control.rabi_rates) == _segment_count
    assert np.allclose(gaussian_control.durations, _duration / _segment_count)

    # check total rotation of pulse
    assert np.isclose(rabi_rotation, _rabi_rotation)
```

qctrlopencontrols/driven_controls/predefined.py
```diff
@@ -973,6 +973,111 @@ def new_wamf1_control(
...
+    )
+
+    check_arguments(
+        width > 0.0,
+        "Width of ideal Gaussian pulse must be greater than zero.",
+        {"width": width},
+    )
+
+    # default mean of the gaussian shaped pulse as a fraction of its duration
+    _pulse_mean = 0.5
```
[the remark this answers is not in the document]
tiborr: Yes :) Should it also be removed in `new_modulated_gaussian_control`?

tests/test_predefined_driven_controls.py
leoadec: Cool, that looks good to me. @tachikoma-li Any further thoughts about the test?

tachikoma-li: LGTM. Nice.

tests/test_predefined_driven_controls.py
```diff
@@ -550,6 +551,52 @@ def test_walsh_control():
...
+    # compute total rotation of generated pulse
+    rabi_rotation = np.dot(gaussian_control.rabi_rates, gaussian_control.durations)
+
+    _segment_width = _duration / _segment_count
+    midpoints = np.linspace(
+        _segment_width / 2, _duration - _segment_width / 2, _segment_count
+    )
+
+    def gauss(t):
+        return np.exp(-0.5 * ((t - _duration / 2) / _width) ** 2)
```
leoadec, line 579: Sorry, I committed your suggested changes to this pull request but forgot that pylint doesn't like single-letter variable names like `t`. Just change `t` to something longer like `time` and I'll be happy to approve your pull requerst.

tiborr:
```suggestion
    def gauss(time):
        return np.exp(-0.5 * ((time - _duration / 2) / _width) ** 2)
```

qctrlopencontrols/driven_controls/predefined.py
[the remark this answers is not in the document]
tachikoma-li: I guess we would focus on the new function in this PR such that the diff/log (part of the reasons we are using squash merge here) is easy to follow and the changes is consistent with the PR title.
Feel free to refactor the code to improve readability in follow-up PRs.

leoadec (approved):
Looks good, thanks again for your contribution

tachikoma-li (approved):
Great work! Thanks for your contribution!

PR #186, opened 2021-05-07
fix: Remove scipy from dependencies

leoadec:
Using `grep`, I couldn't find any use of `scipy` in the repo, so I think we can remove it from the dependencies.

- Remove `scipy` from dependencies in `pyproject.toml`, `poetry.lock`, and `setup.py`.


tachikoma-li (approved):
Thanks Leo!

charmasaur (approved):

PR #195, opened 2021-06-15
fix: Fix path in README

tachikoma-li:
Use correct path to license in README

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -167,4 +167,4 @@ See
 
 ## License
 
-See [LICENSE](LICENSE).
+See [LICENSE](../LICENSE).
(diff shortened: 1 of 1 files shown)


leoadec (commented):
Thanks for this. Just an extra request: I think the link to the "Jupyter notebooks" in the "Usage" section might also need to be updated.

tachikoma-li:
> Thanks for this. Just an extra request: I think the link to the "Jupyter notebooks" in the "Usage" section might also need to be updated.

Good catch. Should be fixed now.

leoadec (approved):
Perfect

charmasaur:
Thanks Li!

PR #200, opened 2021-07-29
feat: Add function to sample a driven control

charmasaur:

PR #201, opened 2021-07-29
feat: Add function to resample a driven control

charmasaur:
Alternative to https://github.com/qctrl/python-open-controls/pull/200. Even though this one is a bit less consistent with BOULDER OPAL, I think it's better. For a start, it means the user can also benefit from resampling for the export-to-file function too, which I would have thought should be very useful. Also, I don't think consistency with BOULDER OPAL is that important here, because DrivenControl is a very different beast to a plain old pulse.

### DIFF
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -20,8 +20,6 @@ import json
 from typing import (
     Dict,
     Optional,
-    Tuple,
-    Union,
 )
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -330,6 +330,51 @@ def duration(self) -> float:
...
+            slightly larger than the original duration, if the time step doesn't exactly divide the
+            original duration.
+        """
+        check_arguments(
+            time_step > 0,
+            "Time step must be positive.",
+            {"time_step": time_step},
+        )
+        check_arguments(
+            time_step <= self.duration,
```
tachikoma-li, line 358: What if the `time_step` is larger than the minimum segment duration time, but smaller than duration? I guess we don't expect it?

charmasaur: Yeah, I was thinking that's just a case we need to accept. Even for a sensible `time_step`, you can still get a wonky sampling if it doesn't exactly divide a segment duration, which I think is unavoidable.

tachikoma-li (approved):
🚀

PR #203, opened 2021-10-15
feat: Add support for Python 3.10

leoadec:

- Make package compatible with Python 3.10.
- Include Python 3.10 in the unit tests.
- Update version of pytest.


qctrlopencontrols/utils.py
```diff
@@ -52,27 +52,27 @@ def create_repr_from_attributes(class_instance=None, attributes=None):
             "Class instance must be a valid object.", {"class_instance": class_instance}
         )
 
-    class_name = "{0.__class__.__name__}".format(class_instance)
+    class_name = f"{class_instance.__class__.__name__}("
```
tachikoma-li, line 55: maybe we don't need `(` at the end?

tachikoma-li (approved):
🎉

cfarrend (approved):

PR #205, opened 2021-11-11
misc: Drop support for Python 3.6

tachikoma-li:


leoadec (dismissed):

perjlieb (dismissed):
🚀

pyproject.toml
```diff
@@ -36,7 +36,7 @@ packages = [
 ]
```
hughhm, line 36: I think line `25` needs to be removed too? 
```
    "Programming Language :: Python :: 3.6",
```

hughhm (changes requested):

pyproject.toml
tachikoma-li: Thanks, @hughhm. 
This line is now removed, also add one entry for 3.9.

hughhm (approved):

leoadec (approved):

PR #206, opened 2021-11-15
misc: Update version of the package pillow in poetry.lock

leoadec:
I received an alert saying that the version of `pillow` in the `poetry.lock` contained a vulnerability (https://github.com/advisories/GHSA-7534-mm45-c74v). This should update it to a version that is no longer vulnerable.

- Update version of pillow in the poetry.lock


tachikoma-li (approved):
Thanks @leoadec !

PR #213, opened 2022-01-07
misc: Update license

tachikoma-li:
Update license. 
We also change the license for `docs/conf.py` from the Q-CTRL one to the Apache one.


leoadec (approved):
Looks good, thanks! I'm going to ping @hughhm as well in case devops is planning to automate the update of the licenses this year, but I think it's fine to go ahead and merge.

hughhm:
Thanks @leoadec, looks like its on our list but not in place just yet

PR #215, opened 2022-01-13
fix: Update pillow

leoadec:
Basically a copy of this dependabot PR: https://github.com/qctrl/python-open-controls/pull/214

Just updating the version of pillow in the `poetry.lock` file.

- Update version of `pillow` in the `poetry.lock` file.


tachikoma-li (approved):
⚡

PR #217, opened 2022-01-13
feat!: Update NumPy version

tachikoma-li:
Update NumPy to 1.20+ (see also  #216).

BREAKING CHANGE:
Users/developers need to upgrade NumPy to 1.20+.


leoadec (approved):
⬆️

PR #219, opened 2022-01-13
fix: Fix type issue with NumPy

tachikoma-li:
To prepare upgrading to NumPy 1.20+, this PR fixes the type issue as type hint can work with most of the NumPy objects (instead of being `Any` before).


leoadec (approved):
📝

PR #222, opened 2022-01-23
misc: Update ipython version to 7.31.1

tachikoma-li:
To address #221. `ipython` is used by `nbval`. It seems `ipython` has dropped support of Python 3.6 since `7.17` ( see [here](https://ipython.readthedocs.io/en/stable/whatsnew/version7.html#ipython-7-17)).
We might just update to 7.31.1 (the latest `7.X` version and `8.X` is still in alpha/beta stage).

- update `ipython` to 7.31.1


leoadec (approved):
Thanks Li!

PR #228, opened 2022-04-28
misc: Tweak release CI

tachikoma-li:
@cfarrend, I am about to do a release for open control and  thinking this might be the chance that we can test whether this would fix the V2 issue. If not, I guess we could downgrade to V1 in other repos.

See related  issues:

https://github.com/actions/checkout/issues/217
https://github.com/actions/checkout/issues/290


cfarrend (approved):
@tachikoma-li for testing you can try your own repo as well, I've done it a few times when it comes to testing GHA.

PR #233, opened 2022-07-21
misc: Add link in the README to the hosted reference docs

leoadec:
- Add information in the README about how to find the reference documentation hosted online.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -85,7 +85,8 @@ pytest
 
 ## Usage
 
-See the [Jupyter notebooks](../examples).
+See the [Jupyter notebooks examples](../examples) and the
+[Q-CTRL Open Controls reference documentation](https://docs.q-ctrl.com/open-controls/references/qctrl-open-controls/).
 
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


perjlieb (approved):
🔗

.github/README.md
```diff
@@ -107,7 +111,7 @@ To build locally:
     poetry run make html
     ```
 
-    If using setuptools:
+    If using pip:
```
tachikoma-li, line 114: I guess `make html` is true for any python environment as long as the virtual env is activated. The above poetry one is to use its own virtual env. But it's fine to make the change here.

tachikoma-li (approved):

.github/README.md
leoadec: Yeah, I was a bit unsure about what to write here. Perhaps "if not using Poetry" or "if using a different/any other Python virtual environment"?

tachikoma-li: Nah, I guess the current change is fine since the whole README emphasize the difference between pip and poetry. Might be better to keep the same here. It just reminds me of the strange issue on M1 recently.

PR #234, opened 2022-07-27
misc: Use positive in error messages

perjlieb:

- Uses positive instead of greater than zero in error messages if value > 0
- Uses non-negative in error messages instead of greater than zero in error messages if value >= 0

### DIFF
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -107,7 +107,7 @@ class DrivenControl:
         # check if all the durations are greater than zero
         check_arguments(
             all(durations > 0),
-            "Duration of driven control segments must all be greater than zero.",
+            "Duration of driven control segments must all be positive.",
             {"durations": durations},
         )
 
@@ -144,10 +144,10 @@ class DrivenControl:
         azimuthal_angles = np.asarray(azimuthal_angles, dtype=float)
         detunings = np.asarray(detunings, dtype=float)
 
-        # check if all the rabi_rates are greater than zero
+        # check if all the rabi_rates are non-negative
         check_arguments(
             all(rabi_rates >= 0.0),
-            "All Rabi rates must be greater than zero.",
+            "All Rabi rates must be non-negative.",
             {"rabi_rates": rabi_rates},
         )
 
--- a/qctrlopencontrols/driven_controls/driven_control.py
+++ b/qctrlopencontrols/driven_controls/driven_control.py
@@ -45,7 +45,7 @@ class DrivenControl:
         of segments.
     rabi_rates : np.ndarray, optional
         The Rabi rates :math:`\{\Omega_n\}` for each segment, in units of radians per second. Every
-        element must be non-negative. Represented as a 1D array of length :math:`N`, where :math:`N`
+        element must be nonnegative. Represented as a 1D array of length :math:`N`, where :math:`N`
         is number of segments. You can omit this field if the Rabi rate is zero on all segments.
     azimuthal_angles : np.ndarray, optional
         The azimuthal angles :math:`\{\phi_n\}` for each segment. Represented as a 1D array of
@@ -144,10 +144,10 @@ class DrivenControl:
         azimuthal_angles = np.asarray(azimuthal_angles, dtype=float)
         detunings = np.asarray(detunings, dtype=float)
 
-        # check if all the rabi_rates are non-negative
+        # check if all the rabi_rates are nonnegative
         check_arguments(
             all(rabi_rates >= 0.0),
-            "All Rabi rates must be non-negative.",
+            "All Rabi rates must be nonnegative.",
             {"rabi_rates": rabi_rates},
         )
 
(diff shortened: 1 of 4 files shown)


qctrlopencontrols/driven_controls/driven_control.py
```diff
@@ -144,10 +144,10 @@ def __init__(
         azimuthal_angles = np.asarray(azimuthal_angles, dtype=float)
         detunings = np.asarray(detunings, dtype=float)
 
-        # check if all the rabi_rates are greater than zero
+        # check if all the rabi_rates are non-negative
         check_arguments(
             all(rabi_rates >= 0.0),
-            "All Rabi rates must be greater than zero.",
+            "All Rabi rates must be non-negative.",
```
jimnel, line 150: Should we instead use nonnegative?
https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/n/non
Or is this a case when the hyphen avoids confusion?

jimnel: Although maybe since we've always used non-negative for error messages we should just use that here and perhaps change it everywhere at a later date.

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -305,12 +305,12 @@ def new_carr_purcell_sequence(
...
     check_arguments(
         duration > 0,
-        "Sequence duration must be greater than zero.",
+        "Sequence duration must be positive.",
         {"duration": duration},
     )
     check_arguments(
         offset_count >= 1,
-        "Number of offsets must be greater than zero.",
+        "Number of offsets must be non-negativei.",
```
jimnel, line 313:
```suggestion
        "Number of offsets must be non-negative.",
```

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -388,12 +388,12 @@ def new_cpmg_sequence(duration, offset_count, pre_post_rotation=False, name=None
...
     check_arguments(
         duration > 0,
-        "Sequence duration must be greater than zero.",
+        "Sequence duration must be positive.",
         {"duration": duration},
     )
     check_arguments(
         offset_count >= 1,
-        "Number of offsets must be greater than zero.",
+        "Number of offsets must be non-negative.",
```
jimnel, line 396:
```suggestion
        "Number of offsets must be positive.",
```

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -467,12 +467,12 @@ def new_uhrig_sequence(duration, offset_count, pre_post_rotation=False, name=Non
...
```
jimnel, line 475:
```suggestion
        "Number of offsets must be positive.",
```

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -546,12 +546,12 @@ def new_periodic_sequence(duration, offset_count, pre_post_rotation=False, name=
...
     check_arguments(
         duration > 0,
-        "Sequence duration must be greater than zero.",
+        "Sequence duration must be positve.",
         {"duration": duration},
     )
     check_arguments(
         offset_count >= 1,
-        "Number of offsets must be greater than zero.",
+        "Number of offsets must be non-negative.",
```
jimnel, line 554:
```suggestion
        "Number of offsets must be positive.",
```

qctrlopencontrols/driven_controls/driven_control.py
perjlieb: Followed the design doc on this, but should be without a hyphen.

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
perjlieb: No,`offset_count` can be zero.

jimnel:
In a few places the condition is 
x >= 1
and you've put
x must be non-negative, but non-negative means it could be zero. It should be positive.

perjlieb:
> In a few places the condition is x >= 1 and you've put x must be non-negative, but non-negative means it could be zero. It should be positive.

That's already resolved. In the first version I only searched for the `>=`, hence the error.

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -893,12 +893,12 @@ def new_x_concatenated_sequence(
...
     check_arguments(
         duration > 0,
-        "Sequence duration must be greater than zero.",
+        "Sequence duration must be positive.",
         {"duration": duration},
     )
     check_arguments(
         concatenation_order >= 1,
-        "Concatenation oder must be greater than zero.",
+        "Concatenation oder must be positive.",
```
jimnel, line 901:
```suggestion
        "Concatenation order must be positive.",
```

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -998,12 +998,12 @@ def new_xy_concatenated_sequence(
...
```
jimnel, line 1006:
```suggestion
        "Concatenation order must be positive.",
```

jimnel (approved):

PR #251, opened 2023-01-02
chore: Update list of project keywords

leoadec:
Follow up to https://github.com/qctrl/template/pull/20

- Update list of keywords in `pyproject.toml` to match the one in the [template repo](https://github.com/qctrl/template/blob/master/pyproject.toml).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,7 @@ documentation = "https://docs.q-ctrl.com/open-controls/references/qctrl-open-con
 keywords = [
     "black opal",
     "boulder opal",
+    "fire opal",
     "nisq",
     "open controls",
     "q control",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -10,9 +10,32 @@ homepage = "https://q-ctrl.com"
 repository = "https://github.com/qctrl/python-open-controls"
 documentation = "https://docs.q-ctrl.com/open-controls/references/qctrl-open-controls/"
 keywords = [
+    "black opal",
+    "boulder opal",
+    "nisq",
```
tachikoma-li, line 15: Do we need to add "fire opal" above?

leoadec: Oh true! I must have copied the old list by mistake. Thanks for spotting the missing line

tachikoma-li (approved):
🪨 🔥

PR #258, opened 2023-02-10
build: Do not show type hint in the signature

tachikoma-li:
See context in https://github.com/qctrl/python-visualizer/pull/266

Also fix a link in the doc.


abenseny (approved):
2️⃣
(thanks for the doi fix!)

tachikoma-li:
> two (thanks for the doi fix!)

Can we do a patch release after this?

PR #261, opened 2023-03-09
docs: Update segment write key to use the prod environment

leoadec:
- Update the segment key used in the reference docs to the prod one, so that Intercom will stop using the "Q-CTRL Test" environment.


abenseny (approved):
🔑

PR #267, opened 2023-04-13
chore!: Follow the naming standards in Open Controls

leoadec:
According to our [naming standards](https://code.q-ctrl.com/naming-conventions#python-packages), package names, install names, and import names that include a product name shouldn't be preceded by "Q-CTRL".

As "Open Controls" is listed as a product in the "Product" tab of the company website, the names in this package have to be updated to remove the initial "Q-CTRL" (except for the project name "Q-CTRL Python Open Controls").

- Rename the package name to "Open Controls" in the README and documentation.
- Update the install name to `open-controls`.
- Update the import name to `opencontrols`.

BREAKING CHANGE: Change import name to `opencontrols`.


docs/conf.py
```diff
@@ -53,9 +53,9 @@
 # Taken from qctrl/docs, see head.html in the repo
 html_context = {
     "var_url": "https://docs.q-ctrl.com",
-    "var_title": "Open Controls Python package | Q-CTRL",
+    "var_title": "Open Controls Python",
```
tachikoma-li, line 56: Do we still need the Python here?

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [tool.poetry]
-name = "qctrl-open-controls"
+name = "open-controls"
 version = "10.0.0"
 description = "Q-CTRL Python Open Controls"
```
tachikoma-li, line 4: Is `description` the same as the project name?
```suggestion
description = "Q-CTRL Open Controls"
```
or 
```suggestion
description = "Open Controls"
```
or

README.md
```diff
@@ -1,4 +1,4 @@
-# Q-CTRL Open Controls
+# Open Controls
 
 Q-CTRL Open Controls is an open-source Python package that makes it easy to
```
tachikoma-li, line 3:
```suggestion
Open Controls is an open-source Python package that makes it easy to
```
Can we also remove Q-CTRL from the rest part of the README?

tachikoma-li (commented):
I guess we also need DevOp to rename this repo?

Not sure how I feel with this breaking change, but maybe this saves a lot of time in some sense.

leoadec:
@tachikoma-li 
> I guess we also need DevOp to rename this repo?

These changes kind of assume no change in the name of the repo. Or, rather, they are agnostic about changes in the name of the repo.

How does this work? Well, we follow the [algorithm here](https://code.q-ctrl.com/naming-conventions) and start with the repo name:

**repo name:** `python-open-controls`

To find the "Project name" we transform the slug into a sentence and add "Q-CTRL" to the front of it:

**Project name:** Q-CTRL Python Open Controls

Note that we're not touching the project name in this PR though, although we can change it later if we want to drop the "Python" part for example.

Now, for the "Package name", we drop the language ("Python") and "Q-CTRL" if it is coming just before the name of a product ("Open Controls"). With this we get:

**Package name:** Open Controls

Once we have the package name, we slugify it to get the install name:

**Install name:** `open-controls`

and then remove the hyphens from the name to obtain the import name:

**Import name:** `opencontrols`

So this is basically just applying what's in the current naming conventions without changing the name of the repo. But if we're not happy with the results we can always change the naming conventions instead.

docs/conf.py
leoadec: Oops, no we don't, that was my oversight. Fixed now

pyproject.toml
leoadec: Yes, this would be the package name, which hasn't changed. If we remove the `python` from the repo name then it would become "Q-CTRL Open Controls" as suggested.

README.md
leoadec: If I remember correctly the only rule is that the title of the `README.md` has to match the Package name. I don't have strong opinions about changing the rest of the text, but I'm happy to do it if you prefer.

tachikoma-li:
> @tachikoma-li
> 
> > I guess we also need DevOp to rename this repo?
> 
> These changes kind of assume no change in the name of the repo. Or, rather, they are agnostic about changes in the name of the repo.

Thanks! These all make sense. I guess one thing still a bit unclear to me is that it seems we begin with a repo name first and then define the project name. Is there any reason that we will keep using `python-open-controls` instead of `open-controls`? Or will that be done in another step?

README.md
tachikoma-li: Not really, I *thought* there is a rule about updating them in all places. If not, I am fine with what you have here.

leoadec:
@tachikoma-li 

> Is there any reason that we will keep using python-open-controls instead of open-controls? Or will that be done in another step?

The latter, I was thinking of the change `python-open-controls` -> `open-controls` as a separate and independent step, which is why it is not covered here. I think we can do it after the renaming of the adapter packages (just in case renaming too many things at once causes some unexpected issue).

leoadec:
Open Controls is no longer listed as a product in the products tab of the website, so these changes no longer apply. So I'm closing this

PR #254, opened 2023-01-23
style: Use the Q-CTRL Sphinx Theme for documentation

leoadec:
This replaces the custom implementation of the RTD Sphinx theme with the [Q-CTRL Sphinx Theme](https://pypi.org/project/qctrl-sphinx-theme), as prescribed in the [code standards for documentation](https://code.q-ctrl.com/documentation#reference-documentation-for-python-packages).

FST available at: https://abc-4000.docs.q-ctrl.com/open-controls/references/qctrl-open-controls

Changes proposed in this pull request:
- Use the Q-CTRL Sphinx Theme for the reference documentation.

PR #276, opened 2023-06-13
build: Remove legacy code from docs/conf.py

leoadec:
There's some legacy code in the file `docs/conf.py` that is no longer necessary to build the reference docs.

As can be seen from the code comment, these were variables to be inserted in the file `_templates/layout.html`, which is no longer present in this repo.

(I removed that file as part of #254, but didn't realize that the new `_templates/layout.html` file from the Q-CTRL Sphinx Theme doesn't need any of those variables.)

- Remove legacy code from `docs/conf.py`.


abenseny (approved):
🪥

PR #315, opened 2025-03-31
ci: Add new reuseable on-push python workflows

jrosco:
Updated CI workflow to use the latest Python on-push resuseable workflows for linting, building, and internal publishing.

This ensures consistency, improves automation, and makes it easier to manage any future changes.

Addresses: https://qctrl.atlassian.net/browse/INFRA-2128

### DIFF
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -8,55 +8,17 @@ on:
       - "*.*"
 
 jobs:
-  housekeeping:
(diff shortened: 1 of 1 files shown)


cfarrend (approved):

cfarrend (changes requested):
Actually, noticed that there's a failure in the workflow config

https://github.com/qctrl/open-controls/actions/runs/14072265887/workflow

PR #318, opened 2025-06-18
ci: Add docker build push docker workflow

jrosco:
Update the docker build ci workflow which now uses the installed `docker
build` command on the workflow runner.

This job only builds an image so there is no need to add a tag with the
suffix `-docs` this is mainly for testing the docker build runs without
any issues.

### DIFF
--- a/.github/workflows/on-pull-request.yaml
+++ b/.github/workflows/on-pull-request.yaml
@@ -34,8 +34,5 @@ jobs:
     - uses: actions/checkout@v4
     - name: Build Sphinx Image
       run: |
-        source <(curl -sL http://ci.q-ctrl.com)
-        ./ci docker build \
-          -f ./docs/Dockerfile \
-          --suffix -docs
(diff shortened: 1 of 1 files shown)


.github/workflows/on-pull-request.yaml
```diff
@@ -34,8 +34,5 @@ jobs:
     - uses: actions/checkout@v4
     - name: Build Sphinx Image
       run: |
-        source <(curl -sL http://ci.q-ctrl.com)
-        ./ci docker build \
-          -f ./docs/Dockerfile \
-          --suffix -docs
```
aggronerd, line 40: @jrosco What does this do and are we going to miss these args?

aggronerd (approved):

.github/workflows/on-pull-request.yaml
jrosco: @aggronerd only builds image, so the `--suffix -docs` seems pointless

PR #334, opened 2026-02-08
feat: Add platonic universal dynamic decoupling sequence for qudits

hectorBrown:
- Implements the platonic dynamical decoupling sequence outlined [here](https://arxiv.org/pdf/2409.04974)  for robust qudit decoupling as a predefined sequence.


tests/test_predefined_dynamical_decoupling.py
```diff
@@ -843,3 +844,117 @@ def test_if_xy_concatenated_sequence_is_identity():
     )
 
     assert _pulses_produce_identity(xy_concat_sequence)
+
+
+def test_platonic_sequence():
```
abenseny: This test ends up being very long and convoluted. If it fails at any point it'd be hard to assess from the failure e.g. for which `order` it broke. I would suggest splitting it into separate tests for the different orders. (We could also use something like `pytest.mark.parametrize`, but I think the code would end up still being quite convoluted with so many options, which we don't want in simple unit tests.)

qctrlopencontrols/dynamic_decoupling_sequences/predefined.py
```diff
@@ -1228,3 +1228,183 @@ def _concatenation_xy(concatenation_sequence) -> np.ndarray:
...
+        "Dihedral": [0, 1, 0, 1, 1, 0, 1, 0],
+        "Tetrahedral": [ 0, 1, 0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 0, 0],
+        "Octahedral": [ 0, 1, 0, 0, 0, 1, 1, 1, 0, 1, 0, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 0, 0, 0, 0, 1, 0, 1, 1, 1, 0, 1, 0, 0, 1, 1, 0, 0, 0, 0, 1, 0, 1, 1, 1, 0, 1, 1 ],
+        "Icosahedral": [ 1, 0, 0, 0, 1, 1, 0, 0, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 1, 0, 1, 1, 1, 0, 1, 0, 0, 1, 1, 0, 0, 1, 1, 0, 1, 1, 0, 1, 1, 0, 1, 1, 1, 0, 0, 0, 0, 1, 0, 1, 1, 1, 0, 0, 0, 1, 0, 1, 1, 1, 0, 0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 0, 0, 1, 1, 0, 0, 1, 1, 0, 0, 1, 1, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 1, 0, 1, 0, 0, 0, 0, 0 ]
+    }
+    # fmt: on
+
+    # The generators assocaited with each point group.
+    phi = (np.sqrt(5) + 1) / 2  # golden ratio
+    generators = {
```
abenseny: Rather than defining these `generators` and `eulerian_paths` dictionaries, can we use`match/case` or an `if-elif` chain and directly build the arrays in the different branches? It would save some memory and make the code more readable. There is no advantage in having an object that holds all the information at once in the implementaiton.

tests/test_predefined_dynamical_decoupling.py
```diff
@@ -843,3 +844,117 @@ def test_if_xy_concatenated_sequence_is_identity():
...
+        _spacing = duration / count
+
+        _offsets = np.array([(_spacing * 0.5 + i * _spacing) for i in range(count)])
+        # fmt: off
+        eulerian_paths = {
+            "Dihedral": [0, 1, 0, 1, 1, 0, 1, 0],
+            "Tetrahedral": [ 0, 1, 0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 0, 0],
+            "Octahedral": [ 0, 1, 0, 0, 0, 1, 1, 1, 0, 1, 0, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 0, 0, 0, 0, 1, 0, 1, 1, 1, 0, 1, 0, 0, 1, 1, 0, 0, 0, 0, 1, 0, 1, 1, 1, 0, 1, 1 ],
+            "Icosahedral": [ 1, 0, 0, 0, 1, 1, 0, 0, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 1, 0, 1, 1, 1, 0, 1, 0, 0, 1, 1, 0, 0, 1, 1, 0, 1, 1, 0, 1, 1, 0, 1, 1, 1, 0, 0, 0, 0, 1, 0, 1, 1, 1, 0, 0, 0, 1, 0, 1, 1, 1, 0, 0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 0, 0, 1, 1, 0, 0, 1, 1, 0, 0, 1, 1, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 1, 1, 0, 1, 0, 0, 0, 0, 0 ]
+        }
```
abenseny: Rather than replicating a lot of the logic of the implementation in the test, we could use hardcoded arrays for the expected values.

abenseny (commented):
Thanks for the contribution @hectorBrown !

Unfortunately, myself and the other maintainers of this package are not in the Sensing division, so I have little to share about their hardware challenges.

hectorBrown:
Hi, appreciate you taking the time to read over this. I've made the changes you suggested above. I also moved from generating the sequences with python generators to numpy for speed and readability. I worry that the tests are a bit unwieldy, especially the Icosahedral one - would it be possible to store these arrays elsewhere?

PR #346, opened 2026-07-22
build!: Update numpy and Python

leoadec:
Update numpy in the lockfile to version 2.4.6, in order to address the security alert https://intel.aikido.dev/cve/AIKIDO-2026-795744

Note that numpy 2.4.x is not available for Python 3.10, so support for the older version of Python had to be dropped.

BREAKING CHANGE: Dropped support for Python 3.10


tachikoma-li (approved):
time flies, we are at 3.14 now..

aggronerd (approved):

PR #347, opened 2026-07-28
ci: Do not persist credentials

abenseny:
Addresses: https://qctrl.atlassian.net/browse/SCUP-5619


baileyivancic (dismissed):
We'll probably want to update the versions on these soon, but happy to leave to another PR

abenseny:
@baileyivancic 
> We'll probably want to update the versions on these soon, but happy to leave to another PR

Yeah, good point. Sorry about that. Should have done it already. I've bumped them.

baileyivancic (approved):

